PR #15, opened 2016-09-21
Story 1970. Renamed LoggerUtil to LogArgUtil. Removed runtime configu…

yeunga:
…ration support from LogArgUtil and associated unused classes.


yeunga:
Travis failed on the jdk7 build. Looks like it failed on the DelegationTokenTest. Wondering if I should generate another pull request to see if it is a transient error.

PR #21, opened 2016-11-24
added support for ArgumentMap to store non-option command-line args

pdowler:
… -) in a list to preserve order


brianmajor:
I've reviewed the code and it looks good, but I don't have permissions to make the merge.

PR #30, opened 2017-02-02
HttpDelete throwing wrong exceptions and wrong status codes.

at88mph:


at88mph:
I agree, but HttpPost does it like that as well.  Are we only changing it here?

at88mph:
Message change updated.

PR #42, opened 2017-09-29
S2158

andamian:


andamian:
I disabled grouping and ordering of imports as I didn't find it particularly useful. Mind you I've had this experience from back in the days of C development but now with the modern IDE it won't be a problem to impose these rules.

PR #48, opened 2018-02-08
Add client IP lookup, and make clients use it.

at88mph:


cadc-log/src/test/java/ca/nrc/cadc/log/WebServiceLogInfoTest.java
```diff
@@ -160,6 +163,62 @@ public void testPathIsJobID()
         EasyMock.verify(request);
     }
 
+    @Test
+    public void testForwardedClientIP()
+    {
```
brianmajor: Checkstyle--I see you have been updating the bracket style so it should be applied here too.

cadc-util/src/main/java/ca/nrc/cadc/xml/XmlUtil.java
```diff
@@ -260,32 +253,50 @@ public static SAXBuilder createBuilder(Map<String, String> schemaMap)
...
+     */
+    public static boolean hasCommentContaining(final Element element, final String commentValue) {
+        return !element.getContent(new ContentFilter(ContentFilter.COMMENT) {
+            @Override
+            public Content filter(final Object obj) {
+                final Content c = super.filter(obj);
+                return ((c != null) && c.getValue().contains(commentValue)) ? c : null;
+            }
+        }).isEmpty();
+    }
```
brianmajor: I find this difficult to read with the ! at the start of the expression and the isEmpty() at the end.  I think it would be more maintainable if it were split into separate lines of logic.

But in general, I wonder if this type of utility is useful in a core library?  The method is quite specific and and I can't really see it being used outside of the integration test for client IP support.  So, I'd suggest moving it to the actual test.

at88mph: It is used in both `cadc-test-vosi` and `cadc-vosi`.  The `XmlUtil` just seemed like a natural place to put it.  Is there a shared Test class?

brianmajor: I don't see it being used in cadc-vosi (and it shouldn't be) because cadc-test-vosi should be the only thing that needs to do this check in the integration tests.  vosi itself doesn't have to read comments from XML.

at88mph: There is a `src/test/java/ca/nrc/cadc/vosi/AvailabilityTest` (unit test) in `cadc-vosi`, and a `src/test/java/ca/nrc/cadc/vosi/AvailabilityTest` (integration test) in `cadc-test-vosi`.  Do we only want to support one of those?

PR #49, opened 2018-03-09
s2248: improve authenticated web service calls

hjeeves:
cadc-util v 1.1.0: supports creating multiple cookies, one for each production domain.


cadc-util/src/main/java/ca/nrc/cadc/auth/AuthenticationUtil.java
```diff
@@ -295,9 +295,11 @@ else if (token != null)
             publicCred.add(token);
             am = AuthMethod.TOKEN;
         }
-        else if (cookie != null)
+        else if (cookieList != null && cookieList.size() != 0)
         {
-            publicCred.add(cookie);
+            for (SSOCookieCredential cookie: cookieList) {
+                publicCred.add(cookie);
+            }
```
brianmajor: You could use the addAll() method here...

cadc-util/src/main/java/ca/nrc/cadc/auth/SSOCookieCredential.java
```diff
@@ -68,10 +72,23 @@ public String getDomain()
...
+        }
+        return expired;
+    }
+
     @Override
     public String toString()
     {
-        return getClass().getSimpleName() + "[" + domain + "," + ssoCookieValue + "]";
+        return getClass().getSimpleName() + "[" + domain + "," + ssoCookieValue
+            + expiryDate.toString() +"]";
```
brianmajor: Expiry date could be null here so that has to be checked.

cadc-util/src/test/java/ca/nrc/cadc/auth/SSOCookieCredentialTest.java
```diff
@@ -0,0 +1,74 @@
...
+    static int MS_IN_HOUR = 3600 * 1000;
+    static
+    {
+        Log4jInit.setLevel("ca.nrc.cadc.auth", Level.INFO);
+    }
+    
+    @Test
+    public void testCookie() throws Exception
+    {
+        Date baseTime = new Date(System.currentTimeMillis());
```
brianmajor: new Date() is the same as new Date(System.currentTimeMillis())

cadc-util/src/test/resources/SupportedDomains.properties
```diff
@@ -0,0 +1,2 @@
+# domains used for generating set of cookies in SSOCookieCredential
+domains = www.canfar.phys.uvic.ca www.cadc.hia.nrc.gc.ca www.ccda.iha.cnrc.gc.ca www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca
```
brianmajor: Maybe mention that the expected value for the domains key is a space separated list of domains.

Also, shouldn't the domains in this list omit the www?  Better ask Dustin about that.

cadc-util/src/main/java/ca/nrc/cadc/auth/SSOCookieCredential.java
```diff
@@ -61,6 +59,19 @@ public SSOCookieCredential(final String cookieValue, final String domain, final
...
+     * expiryDate.
+     * @param cookieValue
+     * @param domain
+     */
+    public SSOCookieCredential(final String cookieValue, final String domain)
+    {
+        this.ssoCookieValue = cookieValue;
+        this.domain = domain;
+        this.expiryDate = null;
+    }
```
brianmajor: This should just call the other constructor so the code isn't in two places

PR #51, opened 2018-04-19
s2247: improve augmentSubject performance by storing more identities in cookies

hjeeves:


cadc-util/src/main/java/ca/nrc/cadc/auth/DelegationToken.java
```diff
@@ -140,27 +194,51 @@ public static String format(DelegationToken token)
...
-        sb.append(token.getUser().getName());
-        if (StringUtil.hasText(token.getUser().getProxyUser()))
+
+        sb.append(EXPIRY_LABEL + VALUE_DELIM);
+        sb.append(token.getExpiryTime().getTime());
+
+        // Add all available identity principals to the content
+        // guarantee order
+        for (Principal prin: token.identityPrincipals) {
+            String tmpKey = "";
```
pdowler: This should use class -> label mapping provided by IdentityType

cadc-util/src/main/java/ca/nrc/cadc/auth/DelegationToken.java
```diff
@@ -140,27 +194,51 @@ public static String format(DelegationToken token)
...
+            if (principalLabels.containsKey(prin.getClass().getSimpleName())) {
+                tmpKey = principalLabels.get(prin.getClass().getSimpleName());
+            } else {
+                continue;
```
pdowler: This "continue" is skipping an unrecognised IdentityType? That should at least be logged (at warn level) or fail. We should discuss impact of that choice.

cadc-util/src/main/java/ca/nrc/cadc/auth/DelegationToken.java
```diff
@@ -190,40 +268,64 @@ public static DelegationToken parse(String text, String requestURI, ScopeValidat
...
         {
             for (String field : fields)
             {
+
                 String key = field.substring(0, field.indexOf(VALUE_DELIM));
                 String value = field.substring(field.indexOf(VALUE_DELIM) + 1);
-                if (key.equalsIgnoreCase("userid"))
+                log.debug("key = value: " + key + "=" + value);
+
+                if (key.equalsIgnoreCase(principalLabels.get(HttpPrincipal.class.getSimpleName())))
```
pdowler: It could be true that all Principal types have a constructor that takes the name as a string... could this use reflection to find and call that ctor? Otherwise, support for new types has to be added in IdentityType and in here...

And other places unless... maybe a static factory method in IdentityType is feasible? 
``Principal IdentityType.create(String name, IdentityType type)``

cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
```diff
@@ -778,6 +778,15 @@ protected void setRequestSSOCookie(HttpURLConnection conn)
                 {
                     if (conn.getURL().getHost().endsWith(cookieCred.getDomain()))
                     {
+                        // HACK ("Pat Said") - this is rather horrenous, but in the java HTTP
+                        // library, the cookie isn't sent with the redirect. But it doesn't flag it
+                        // as a problem. This flags the problem early, allows us to detect attempts
+                        // to send cookies + redirect via POST.
+                        // GET (HttpDownload) works, and sends the cookies as expected.
+                        if (followRedirects && "POST".equals(conn.getRequestMethod())) {
+                            throw new UnsupportedOperationException("Attempt to follow redirect with cookies.");
```
pdowler: I would add (POST) to the end of the message

cadc-util/src/test/java/ca/nrc/cadc/auth/SSOCookieManagerTest.java
```diff
@@ -79,30 +84,58 @@ public void roundTrip() throws Exception
...
-        DelegationToken cookieToken = cm.parse(cm.generate(userPrincipal));
+        DelegationToken cookieToken = cm.parse(cm.generate(userPrincipal, null));
         HttpPrincipal actualPrincipal = cookieToken.getUser();
-        
+
+        //Check principal
         assertEquals(userPrincipal, actualPrincipal);
     }
 
+    public String createCookieString() throws InvalidKeyException, IOException {
```
pdowler: This support methdo looks like it re-implements code in DelegationToken... It is going to every brittle.

I think a couple of tests like testRountTrip() -- testRoundTripMin() and testRoundTripAll() - with 1 identity (userid) and all supported identity types would be better.

cadc-util/src/main/java/ca/nrc/cadc/auth/IdentityType.java
```diff
@@ -112,5 +129,18 @@ public String toString()
         return this.getClass().getSimpleName() + "[" + value + "]";
     }
 
+
+    public static IdentityType get(String identity) {
```
pdowler: Isn't this get(String) duplicating toValue(String)? In general enums are small enough to not need a map to implement so the toValue impl is fine.

cadc-util/src/main/java/ca/nrc/cadc/auth/DelegationToken.java
```diff
@@ -190,40 +266,70 @@ public static DelegationToken parse(String text, String requestURI, ScopeValidat
...
                 String value = field.substring(field.indexOf(VALUE_DELIM) + 1);
-                if (key.equalsIgnoreCase("userid"))
+                log.debug("key = value: " + key + "=" + value);
+
+                if (key.equalsIgnoreCase(IdentityType.USERID.getValue()))
                 {
                     userid = value;
                 }
-                if (key.equalsIgnoreCase("proxyuser"))
+                if (key.equalsIgnoreCase(PROXY_LABEL))
```
pdowler: sequence of if should be "else if"?

cadc-util/src/main/java/ca/nrc/cadc/auth/SSOCookieManager.java
```diff
@@ -161,28 +230,20 @@ public void setOffsetExpiryHours(int offsetExpiryHours)
...
-     * @param expiryDate
      * @return cookieList
      */
-     public List<SSOCookieCredential> getSSOCookieCredentials(final String cookieValue, final String requestedDomain, final Date expiryDate) {
-        List<SSOCookieCredential> cookieList = new ArrayList<>();
-
-        // Make cookie with requested domain
-        SSOCookieCredential requestedCookie = new SSOCookieCredential(cookieValue, requestedDomain, expiryDate);
-        cookieList.add(requestedCookie);
+    public List<SSOCookieCredential> getSSOCookieCredentials(final String cookieValue, final String requestedDomain)
```
pdowler: I find the requestedDomain argument here misleading...

hjeeves: agreed. done

cadc-util/src/main/java/ca/nrc/cadc/auth/DelegationToken.java
hjeeves: switch on strings is available as of Java 7. At the very least I inserted the elsif to reduce the amount of wasted calls.

PR #53, opened 2018-05-02
augment cadc-rest so it is sufficient to implement UWS async

pdowler:
new usage in https://github.com/pdowler/uws/tree/master/cadc-uws-server 

existing usage in https://github.com/pdowler/caom2db/tree/master/caom2-repo-server has been adapted

### DIFF
--- a/cadc-rest/build.gradle
+++ b/cadc-rest/build.gradle
@@ -14,7 +14,7 @@ sourceCompatibility = 1.7
 
 group = 'org.opencadc'
 
-version = '1.1.1'
+version = '1.1.4'
 
 dependencies {
     compile 'commons-fileupload:commons-fileupload:1.3.+'
--- a/cadc-rest/build.gradle
+++ b/cadc-rest/build.gradle
@@ -14,7 +14,7 @@ sourceCompatibility = 1.7
 
 group = 'org.opencadc'
 
-version = '1.1.4'
+version = '1.2'
 
 dependencies {
     compile 'commons-fileupload:commons-fileupload:1.3.+'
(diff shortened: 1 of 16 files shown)


cadc-rest/build.gradle
```diff
@@ -14,7 +14,7 @@ sourceCompatibility = 1.7
 
 group = 'org.opencadc'
 
-version = '1.1.1'
+version = '1.1.4'
```
brianmajor: why the jump from 1.1.1 to 1.1.4?

cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java
```diff
@@ -115,6 +115,8 @@ public void init(ServletConfig config) throws ServletException
         this.putAction = loadAction(config, "put");
         this.deleteAction = loadAction(config, "delete");
         this.headAction = loadAction(config, "head");
+        this.appName = config.getServletContext().getServletContextName();
+        this.restEndpoint = config.getServletContext().getServletContextName()  + "." + config.getServletName();
```
brianmajor: Shouldn't this be separated by a slash instead of a period?

cadc-rest/build.gradle
pdowler: good question, I probably restarted work 3 times :-) Do you think 1.1.2 or 1.2 is more approp?

cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java
pdowler: I really didn't want them to look like paths since they are not - these names are independent of servlet-mappings - they are just named components.

cadc-rest/build.gradle
brianmajor: Probably 1.2

cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java
pdowler: renaming restEndpoint to componentID

PR #54, opened 2018-05-11
rt 75374: added setPath function to WebServiceLogInfo as it's not alw…

hjeeves:
…ays part of information used in constructors, and it's needed for DataTransferServlet to be able to report target path in END log statements.


hjeeves:
Also needing to be reviewed with this is a change in service.git/data, rt75374 branch.

PR #57, opened 2018-05-25
Fixes https://github.com/opencadc/ac/issues/50

at88mph:
The tests were updated to reflect this RFC:
https://tools.ietf.org/html/rfc6265

The identities are encoded and placed in an `identities=` key in the token value.

Rather than the identities being placed directly as before:
expirytime=`EXPIRYTIME`&X500=`DISTINGUISHED_NAME_IDENTITY`&numericID=`NUMERIC_IDENTITY`&userID=`HTTP_USERNAME_IDENTITY`&scope=`SCOPE`&domain=`DOMAINS[]`&signature=`ENCODED_SIGNATURE`

The value now looks like:
expirytime=`EXPIRYTIME`&identities=`ENCODED_SET_OF_IDENTITIES`&scope=`SCOPE`&domain=`DOMAINS[]`&signature=`ENCODED_SIGNATURE`

### DIFF
--- a/cadc-util/build.gradle
+++ b/cadc-util/build.gradle
@@ -2,7 +2,7 @@ plugins {
     id 'java'
     id 'maven'
     id 'maven-publish'
-    id 'com.jfrog.bintray' version '1.7.1'
+    id 'com.jfrog.bintray' version '1.8.0'
 }
 
 repositories {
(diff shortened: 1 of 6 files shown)


at88mph:
The `/whoami` call will fail, and they will need to login again.

The cookie format had to drastically change, I'm not sure how else to implement it.  Do we leave in the existing validation for those existing cookies?

at88mph:
As discussed, the `DelegationToken` will format a value as `$ENCODING:$TOKEN`:
`e.g. base64:98ASDFSDFN2W3398UZxcasd`

The only issue is that the signature is also encoded, as the bare value from the `Signature` class cannot be transported due to odd characters.

PR #59, opened 2018-06-08
improve default path logging in cadc-rest

pdowler:


cadc-http-client/src/main/java/ca/nrc/cadc/net/Main.java
```diff
@@ -111,7 +112,10 @@ else if (am.isSet("v") || am.isSet("verbose"))
             Log4jInit.setLevel("ca.nrc.cadc.net", level);
             Log4jInit.setLevel("ca.nrc.cadc.auth", level);
             
-            Subject s = CertCmdArgUtil.initSubject(am);
+            Subject s = AuthenticationUtil.getAnonSubject();
+            if (am.isSet("cert")) {
+                s = CertCmdArgUtil.initSubject(am);
+            }
```
brianmajor: version number increment for cadc-http-client in build.gradle?

pdowler: It is only a toy command-line that we build locally for testing

PR #61, opened 2018-09-04
Story 2325. Updated JsonInputter and JsonOutputter 

yeunga:
The updates were to support JsonReader and JsonWriter required by the doi service in story 2325. The changes rendered this new version incompatible with the old version and hence the major version has been incremented in build.gradle.


pdowler:
It would be good to include an assessment of the downstream effects:

Can you try running the JsonWriterTest in caom2.git/caom2 and the JsonNodeWriterTest in vos.git/cadc-vos? I think they should pass without change since they don't assert much beyond validity. 

There are also integration tests for caom2ops (MetaIntTest) and vospace (NodeOutputFormatTest).

That won't address the change in output that may effect UI applications (I think restricted to storage).

yeunga:
JsonWriterTest in caom2.git/caom2 passed. 

JsonNodeWriterTest in vos.git/cadc-vos failed. Updated JsonNodeWriterTest and now it passes locally on my desktop. Here is the pull request for the change: https://github.com/opencadc/vos/pull/57 Travis fails because it depends on the new cadc-util (1.2.0) which has not been released yet.

MetaIntTest in caom2ops integration passed.

NodeOutputFormatTest in vospace integration test passed.

The build.gradle files for caom2ops and vospace currently use cadc-util 1.1. For these services to use the new cadc-util, they would need to be changed to use cadc-util 1.2.

yeunga:
Fixed opencadc/ac/cadc-access-control unit tests failures. Issued a pull request (https://github.com/opencadc/ac/pull/53).

PR #62, opened 2018-09-05
 Added runID to logInfo output

brianmajor:


cadc-log/src/main/java/ca/nrc/cadc/log/ServletLogInfo.java
```diff
@@ -121,6 +121,7 @@ public ServletLogInfo(HttpServletRequest request, String path, boolean pathIsJob
         } else {
             this.path = path;
         }
+        runID = request.getParameter("runID");
```
pdowler: param names are expected to not be case-sensitive but iirc the servlet API is... you may need to iterate through to figure out what the caller did

PR #64, opened 2018-09-14
S2325 Added an overloaded constructor to JsonOutputter

yeunga:
This change is to facilitate the DoiJsonWriter in the doi project.


cadc-util/src/main/java/ca/nrc/cadc/xml/JsonOutputter.java
```diff
@@ -141,13 +141,33 @@ public void setFormat(Format fmt)
...
+    }
+
+    /**
+     * Write a Document instance to a writer.
+     *
+     * @param element document to write starting from element.
+     * @param writer Writer to write to.
+     * @throws IOException if the writer fails to write.
+     */
+    public void output(final Element element, Writer writer)
```
pdowler: Why does this have to be public? Some other code can write an element without a document?

PR #68, opened 2018-10-18
HttpPost supports FileContent for submitting data

hjeeves:
Previously, it only supported Files

### DIFF
--- a/cadc-util/src/main/java/ca/nrc/cadc/net/FileContent.java
+++ b/cadc-util/src/main/java/ca/nrc/cadc/net/FileContent.java
@@ -1,0 +1,31 @@
+package ca.nrc.cadc.net;
+
+/**
+ * Created by jeevesh on 2018-10-17.
+ */
+public class FileContent {
+
+    String content;
+    String contentType;
+
+    public FileContent(String content) {
+        this.content = content;
+    }
+
+    public String getContent() {
+        return content;
+    }
+
+    public void setContent(String content) {
+        this.content = content;
+    }
+
+    public String getContentType() {
+        return contentType;
+    }
+
+    public void setContentType(String contentType) {
+        this.contentType = contentType;
+    }
+
+}
--- a/cadc-util/src/main/java/ca/nrc/cadc/net/FileContent.java
+++ b/cadc-util/src/main/java/ca/nrc/cadc/net/FileContent.java
@@ -1,3 +1,72 @@
+/*
+************************************************************************
+*******************  CANADIAN ASTRONOMY DATA CENTRE  *******************
+**************  CENTRE CANADIEN DE DONNÉES ASTRONOMIQUES  **************
+*
+*  (c) 2018.                            (c) 2018.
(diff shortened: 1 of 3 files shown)


hjeeves:
service/tap integration tests work with this version of cadc-util. (My understanding is that is one of the main functions affected by these changes.)

cadc-util/src/main/java/ca/nrc/cadc/net/FileContent.java
```diff
@@ -0,0 +1,31 @@
...
+public class FileContent {
+
+    String content;
+    String contentType;
+
+    public FileContent(String content) {
+        this.content = content;
+    }
+
+    public String getContent() {
```
pdowler: For a class like this I would not have a setContent method (immutable) and I would throw an IllegalArgumentException from the constructor if the content is null. Then the content field can be private and final.

Generally, if a field is allowed to be null it might as well be public and have no methods (contentType).

cadc-util/src/main/java/ca/nrc/cadc/net/HttpPost.java
```diff
@@ -540,6 +553,26 @@ private void writeFilePart(String fieldName, File uploadFile, Writer w)
...
+    {
+        StringBuilder sb = new StringBuilder();
+        log.warn(uploadContent.getContentType());
+        sb.append(LINE_FEED).append("--" + MULTIPART_BOUNDARY);
+        // 'filename' for data entry is needed so this data is treated as
+        // stream input by the accepting web service
+        sb.append(LINE_FEED).append("Content-Disposition: form-data; name=\"" + fieldName + "\";"
+            + " filename=\"dummyFile\"");
+
+        sb.append(LINE_FEED).append("Content-Type: " + uploadContent.getContentType() + ";");
```
pdowler: This line should only be added if contentType is not null; it could otherwise fail in an unpredictable way.

cadc-util/src/main/java/ca/nrc/cadc/net/FileContent.java
hjeeves: I don't know which way to take the FileContent class then. Both fields mandatory? contentType not mandatory? That's the question that needs answering before I can put the code one way or the other.

I set fileContent up to have both be mandatory. Then probably made it unclear by putting a check on it's value in the HttpPost code that prints out the multipart headers?

brianmajor: Both are mandatory, so:

1. Check for null values in constructor
2. Make the attributes private
3. Have a getter for each but no setter

hjeeves: For the code that is dependent on this change, the content type is mandatory. (It's used to filter the stream input to one reader or another. I'm not sure what the requirement would be for other code (ie potentially the "tap" services) that would use this code path? 

I can make contentType optional for FileContent, then put a check in the InlineContentHandler...

PR #72, opened 2018-12-17
support POST of binary content and config of RestAction

pdowler:
cadc-util: add binary content support to HttpPost/FileContent
cadc-rest: pass init-params not used to configure actions into base RestAction

### DIFF
--- a/cadc-util/src/main/java/ca/nrc/cadc/net/HttpPost.java
+++ b/cadc-util/src/main/java/ca/nrc/cadc/net/HttpPost.java
@@ -82,14 +82,17 @@ import javax.net.ssl.HttpsURLConnection;
 import org.apache.log4j.Logger;
 
 import ca.nrc.cadc.util.StringUtil;
+import java.io.BufferedOutputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
+import java.io.FileInputStream;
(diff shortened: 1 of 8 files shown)


cadc-util/src/main/java/ca/nrc/cadc/net/HttpPost.java
```diff
@@ -425,30 +426,33 @@ private void doPost(HttpURLConnection conn, Map<String,List<Object>> params, Map
...
-                if (up.getValue() instanceof File) {
-                    writeFilePart(up.getKey(), ((File)up.getValue()), writer);
-                }
-                else if (up.getValue() instanceof FileContent) {
-                    writeFilePart(up.getKey(), (FileContent)up.getValue(), writer);
-                }
-                else {
-                    throw new UnsupportedOperationException("Unexpected upload type: " + up.getClass().getName());
+                for (Map.Entry<String,Object> up : uploads.entrySet())
+                {
```
brianmajor: May as well update the codestyle here so that open brackets are on the same line as the condition.

pdowler: I have been reluctant to do that and hide  real changes. Maybe delay that and have a with just checkstyle fix?

PR #75, opened 2018-12-19
Accept client certificates from HTTP Header X-Client-Certificate

brianmajor:


cadc-util/src/main/java/ca/nrc/cadc/auth/ServletPrincipalExtractor.java
```diff
@@ -117,10 +123,34 @@ private ServletPrincipalExtractor()
...
-        if (!ArrayUtil.isEmpty(ca)) {
+        
+        if (ArrayUtil.isEmpty(ca)) {
+            // try the header field
+            String certString = req.getHeader(CERT_HEADER_FIELD);
+            log.debug("Cert String: " + certString);
+            if (certString != null && certString.length() > 0) {
+                try {
+                    byte[] certBytes = Base64.decode(certString);
+                    if (certBytes != null) {
```
pdowler: This code should just use SSLUtil.getCertifcates(byte[]) which does Base64 decoding and then SSLUtil.readCertificateChain(byte[]) which does the rest.

PR #83, opened 2019-03-06
Fixes #58

### LINKED ISSUE #58: ca.nrc.cadc.rest.InlineContentHandler does not handle Multipart Request
In order to use the new `cadc-uws-server:1.1`, an application must use the `ca.nrc.cadc.rest.InlineContentHandler`, but it doesn't handle an HTTP Request POSTed with Multipart encoding.  Instead, it assumes the HTTP Request parameters are sufficient.

This breaks existing applications that use File Upload, such as [CADC AdvancedSearch](http://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/en/search/).

at88mph:
Support multiple parameters with the same name for a Multipart request.


cadc-rest/src/main/java/ca/nrc/cadc/rest/SyncInput.java
```diff
@@ -239,107 +237,100 @@ public String getParameter(String name)
...
+        throws FileUploadException, IOException, ResourceNotFoundException {
+        while (itemIterator.hasNext()) {
             FileItemStream item = itemIterator.next();
             String name = item.getFieldName();
             InputStream stream = item.openStream();
-            if (item.isFormField())
-                processParameter(name, new String[] { Streams.asString(stream) });
-            else
+            if (item.isFormField()) {
+                final List<String> existingItems = this.params.get(name);
```
pdowler: Why not just add the new value to the existing list? You could just have processParameter do this much more simply by getting the current list from the params map, create and put if null, and add the value(s).

cadc-rest/src/main/java/ca/nrc/cadc/rest/InlineContentHandler.java
```diff
@@ -101,7 +101,7 @@
     Content accept(String name, String contentType, InputStream inputStream)
         throws InlineContentException, IOException, ResourceNotFoundException;
 
-    public static class Content
+    class Content
```
pdowler: This should be static (so it is clear that lifecycle goes beyond that if the InlineContentHandler that created it. 

And because it is something that implementations create with new it should be marked public. I know that technically works with package access but public reflects the intent better. It also allows one to write an InlineContentHandler impl that goes beyond one class (I think only the class that "implements InlineContentHandler" can see this otherwise.

at88mph: Inner Classes of interfaces are `public` `static` already.

pdowler: OK.

cadc-rest/src/main/java/ca/nrc/cadc/rest/SyncInput.java
```diff
@@ -349,26 +337,21 @@ private void processStream(String name, String contentType, InputStream inputStr
...
-        public CloseWrapper(InputStream in) {
+        final String tname;
+
+        CloseWrapper(InputStream in) {
             super(in);
             this.tname = "CloseWrapper[" + Thread.currentThread().getName() + "] ";
         }
 
         @Override
         public void close() throws IOException {
```
pdowler: Could you leave these 3 lines in but comment them all out? It took several iterations to figure out what info was needed to debug so I don't want to lose them. thnx

at88mph: Sure.

cadc-rest/build.gradle
```diff
@@ -24,4 +24,7 @@ dependencies {
     compile 'org.opencadc:cadc-log:1.+'
 
     testCompile 'junit:junit:4.+'
+    testCompile 'org.mockito:mockito-core:2.+'
```
pdowler: I didn't notice this earlier.... is easymock dead? or no longer easy? :-)

I do not want mock testing used in these libraries - the tests end up too  brittle to be useful. Please remove.

at88mph: EasyMock uses `cglib`, which causes a whack of warnings regarding reflection, so it's been advised to move away from it.

Do we have a better way to test the `HttpServletRequest` requirements?

at88mph: Added stub class for `HttpServletRequest`.

cadc-rest/src/main/java/ca/nrc/cadc/rest/SyncInput.java
```diff
@@ -239,107 +237,97 @@ public String getParameter(String name)
...
                 processStream(null, contentType, request.getInputStream());
             }
         }
     }
 
-    private void processParameter(String name, String[] values)
-    {
-    	this.params.put(name, Arrays.asList(values));
+    private void processParameter(String name, List<String> values) {
+        this.params.put(name, values);
```
pdowler: I would put the code to get the existing List from the params map here, eg:
List<String> cur = params.get(name);
```if (cur == null) 
  params.put(name, values);
else 
 cur.addAll(values);```

and then simplify the processMultipart. It is more efficient and changes to init() or processMultipart are less likely to re-introduce the bug.

at88mph:
Mocks were removed already, see above comment.

cadc-rest/build.gradle
pdowler: The real tests that are useful would be a dummy set of actions and testing in one or more servlet containers. Trying to mock or stub it out just adds another "servlet API impl" that may not be the same as tomcat or resin or whatever. That's why I don't think it is worth the effort to try to test code like this that is just a facade on the servlet API.

at88mph:
Tests removed.

PR #90, opened 2019-05-27
S2453 - Remove HTTP attribute case sensitivity from RestletPrincipalExtractor

andamian:


cadc-auth-restlet/src/test/java/ca/nrc/cadc/auth/restlet/RestletPrincipalExtractorTest.java
```diff
@@ -63,6 +64,54 @@
...
+        reset(getMockRequest());
+        attributes.clear();
+        form = new Form(AuthenticationUtil.AUTH_HEADER.toLowerCase() + "=foo");
+        attributes.put("org.restlet.http.headers", form);
+        expect(getMockRequest().getAttributes()).andReturn(attributes).atLeastOnce();
+        
+		replay(getMockRequest());
+		try {
+				DelegationToken dt = getTestSubject().getDelegationToken();
+				assertTrue(false);
```
pdowler: I prefer Assert.fail("reason: ...") after a call that is supposed to throw.

cadc-auth-restlet/src/test/java/ca/nrc/cadc/auth/restlet/RestletPrincipalExtractorTest.java
```diff
@@ -63,6 +64,54 @@
...
+		} catch (NotAuthenticatedException e)
+		{
+				assertEquals("Unexpected exception", "invalid delegation token. null", e.getMessage());
```
pdowler: The message needs to say "expected exception" so it is clear why it is being caught and dealt with in this way.

PR #94, opened 2019-08-07
Add BearerTokenPrincipal

cbanek:
Add a BearerTokenPrincipal, which is just the authorization header
set with something that starts with Bearer.  This is automatically
pulled out in the ServletPrincipalExtractor and presented as a
Principal on the subject.

### DIFF
--- a/cadc-util/src/main/java/ca/nrc/cadc/auth/BearerTokenPrincipal.java
+++ b/cadc-util/src/main/java/ca/nrc/cadc/auth/BearerTokenPrincipal.java
@@ -1,0 +1,144 @@
+/*
+ ************************************************************************
+ *******************  CANADIAN ASTRONOMY DATA CENTRE  *******************
+ **************  CENTRE CANADIEN DE DONNÉES ASTRONOMIQUES  **************
+ *
+ *  (c) 2019.                            (c) 2019.
+ *  Government of Canada                 Gouvernement du Canada
+ *  National Research Council            Conseil national de recherches
+ *  Ottawa, Canada, K1A 0R6              Ottawa, Canada, K1A 0R6
+ *  All rights reserved                  Tous droits réservés
+ *
+ *  NRC disclaims any warranties,        Le CNRC dénie toute garantie
+ *  expressed, implied, or               énoncée, implicite ou légale,
+ *  statutory, of any kind with          de quelque nature que ce
+ *  respect to the software,             soit, concernant le logiciel,
+ *  including without limitation         y compris sans restriction
+ *  any warranty of merchantability      toute garantie de valeur
+ *  or fitness for a particular          marchande ou de pertinence
+ *  purpose. NRC shall not be            pour un usage particulier.
+ *  liable in any event for any          Le CNRC ne pourra en aucun cas
+ *  damages, whether direct or           être tenu responsable de tout
+ *  indirect, special or general,        dommage, direct ou indirect,
+ *  consequential or incidental,         particulier ou général,
+ *  arising from the use of the          accessoire ou fortuit, résultant
+ *  software.  Neither the name          de l'utilisation du logiciel. Ni
+ *  of the National Research             le nom du Conseil National de
+ *  Council of Canada nor the            Recherches du Canada ni les noms
+ *  names of its contributors may        de ses  participants ne peuvent
+ *  be used to endorse or promote        être utilisés pour approuver ou
+ *  products derived from this           promouvoir les produits dérivés
+ *  software without specific prior      de ce logiciel sans autorisation
+ *  written permission.                  préalable et particulière
+ *                                       par écrit.
+ *
+ *  This file is part of the             Ce fichier fait partie du projet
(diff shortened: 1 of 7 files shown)


cbanek:
Excellent - totally agree on tests, I just wanted to check myself before I get too far down the rabbit hole.  Version number is also not a problem at all!

pdowler:
If you look at the cookie handling you see that it conveys a username and after parsing it creates an HttpPrincipal (yeah - poorly named because we started with HTTP Auth which is username/password). I was expecting oauth to be somewhat like this... 

What is the identity that  is conveyed in a bearer token?

cbanek:
Sadly the bearer token is pretty much opaque.  It just means if you have it, that's all you need.  So there's not really a great way to know what kind of format it is, although in my case, it's a JWT token, which I parse out in my AuthorizerImpl over here:  https://github.com/lsst-sqre/lsst-tap-service/pull/27

This is where I'm taking the BearerTokenPrincipal and changing it over to a set of principals with the info in them.  But there isn't a way to say when you have a token in the auth header what type it is as far as I know.  So that's why I pretty much just stashed the token, and am doing the parsing somewhere else.

If I could parse out the token in the PrincipalExtractor, I think that would make more sense, but that would probably put a little too much of my LSST business logic in that mostly generic layer of code.  Even if we knew it was a JWT token, the claims and fields in it are just a JSON dict and are not necessarily standardized over all JWT tokens, other than that they are also JSON dicts.

I think another way of doing this might be to extend the PrincipalExtractor to allow for me to add maybe some delegates, like, a function to add a new object to call an extract function on.  This would allow me to insert myself into the ServletPrincipalExtractor, which seems to not be as extensible as the AuthorizerImpl (although we could do that).

pdowler:
So bearer token is an opaque object that can be turned into identity(ies) by calling back to a service. In that case I think you are on the right track by just capturing it in the PrincipalExtractor.

The next thing would be to implement your own AuthenticatorImpl which takes the Subject with the token in it and "augment" it with real valid identity(ies). This is where you would call back to your auth-support service and is the intended plugin. Note that while our impl just adds more principals, yours could (as necessary): 
* create and return a new Subject instance
* remove the BearerTokenPrincipal if it is no longer useful as a temp identity
* store the bearer token in the credentials if it could be re-used for other calls (might need some sort of wrapper code like cookies to limit scope where it can/should be used)

cadc-util/src/main/java/ca/nrc/cadc/auth/BearerTokenPrincipal.java
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @return the name of this principal.
+     */
+    @Override
+    public String getName()
+    {
+        return getToken();
+    }
+
+    public String getToken()
```
pdowler: I probably would not bother with getToken() and just make the javadoc for getName() clear that it returns the entire token.

cadc-util/src/main/java/ca/nrc/cadc/auth/ServletPrincipalExtractor.java
```diff
@@ -164,6 +164,12 @@ public ServletPrincipalExtractor(final HttpServletRequest req)
             finally { }
         }
 
+        String authToken = request.getHeader("Authorization");
```
pdowler: I would declare a constant like AuthenticationUtil.HTTP_AUTHORIZATION_HEADER with this value and document it as being defined in HTTP/1.1

cbanek:
Yep I think I've done that in the PR on my side in lsst-sqre/lsst-tap-service#27 although I don't have to call back to a service because everything is encoded in the token.  But I am making a NumericPrincipal (with the UID, although it's a bit strange that it's a UUID, but I think you're just using it as a packing method), and an HttpPrincipal with the username for it.  I'm also keeping around the original BearerTokenPrincipal because if anything needs to call out, then I could use that token on those calls to do impersonation (as long as the tokens are valid).

cadc-util/src/main/java/ca/nrc/cadc/auth/BearerTokenPrincipal.java
```diff
@@ -0,0 +1,115 @@
+/*
```
pdowler: This isn't the full header comment that includes the license terms. You can find the correct one in AuthenticationUtil.

pdowler:
In that case, I would remove the BearerTokenPrincipal from the principals set and store the token in the credentials set for later re-use. Cookies (tokens) are wrapped and include a domain where they chould be sent; we also need to do that with certificate chains and I assume bearer tokens as well. In the past we assumed a single cert was usable everywhere (IVOA SSO-1.x) but that dream seems to have died. BTW, proxy certs get created/stored in the subject's credential set elsewhere (cadc-cdp, CredUtil) but the cookie example should be there.

cbanek:
Okay, pushed a new commit with changes from feedback, added some simple tests, and bumped the version number!  Ready for round 2.

pdowler:
published cadc-util-1.2.18 to jcenter maven repository

PR #97, opened 2019-09-11
Added PosixPrincipal 

yeunga:


cadc-util/src/main/java/ca/nrc/cadc/auth/PosixPrincipal.java
```diff
@@ -0,0 +1,98 @@
...
+	private static final long serialVersionUID = 5423257890488724644L;
+	private int uidNumber;
+
+    /**
+     * Ctor
+     * @param uidNumber unique identifier
+     */
+    public PosixPrincipal(int uidNumber)
+    {
+        this.uidNumber = uidNumber;
```
brianmajor: Maybe check to make sure this is a number > 0

cadc-util/src/main/java/ca/nrc/cadc/auth/PosixPrincipal.java
```diff
@@ -43,60 +43,65 @@
...
+    private int uidNumber;
 
     /**
      * Ctor
      * @param uidNumber unique identifier
      */
-    public PosixPrincipal(int uidNumber)
-    {
+    public PosixPrincipal(int uidNumber) {
     	if (uidNumber > 0) {
```
brianmajor: Sorry, changed my mind here.  I guess we should allow a uidNumber of zero (but not less than zero) in case someone wants to represent the root user.

And, how about formatting the check like:
```
if (uidNumber <= 0) {
    throw new IllegalArg...
}
this.uidNumber = uidNumber
```
then there's no need for the 'else'

PR #106, opened 2019-10-31
S2526 - new json format for info logging

jburke-cadc:


cadc-log/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
```diff
@@ -135,14 +130,50 @@ public String start() {
...
+    private String getServiceName() {
+        return "\"service\":{\"name\":\"" + serviceName + "\"}";
+    }
+
+    private String getThreadName() {
+        return "\"thread\":{\"name\":\"" + Thread.currentThread().getName() + "\"}";
+    }
+
+    private String getLoglevel() {
+        return "\"log\":{\"level\":\"info\"}";
```
brianmajor: Shouldn't we be returning the actual current log level?

cadc-rest/build.gradle
```diff
@@ -14,14 +14,20 @@ sourceCompatibility = 1.7
...
+    compile 'org.opencadc:cadc-log:[1.1.0,)'
 
     testCompile 'junit:junit:4.+'
 }
+// alinga-- 
+jar {
+    from('src/main/java') {
+        include '**/*.java'
+    }
+}
```
brianmajor: This should be removed?

cadc-log/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
```diff
@@ -279,4 +310,18 @@ public void setRunID(String runID) {
         this.runID = runID;
     }
 
+    public void setClass(Class clz) {
+        this.clz = clz;
```
brianmajor: Could the calling class not be determined from the call stack in start() and end()?

cadc-log/src/main/java/ca/nrc/cadc/log/ServletLogInfo.java
```diff
@@ -83,65 +86,38 @@
...
-        Enumeration<String> paramNames = request.getParameterNames();
-        while (paramNames.hasMoreElements()) {
-            String nextName = paramNames.nextElement();
-            if (nextName.equalsIgnoreCase("runid")) {
-                runID = request.getParameter(nextName);
-                break;
-            }
+        String pathInfo = request.getPathInfo();
+        if (pathInfo == null) {
+            pathInfo = "";
```
brianmajor: runID is no longer parsed out here.  How is it handled now?

cadc-log/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
jburke-cadc: This class is used only for info level logging. You could add a logging level variable to the class, default it to info, and add a function to set the log level. But it would, or should, always be at the info level.

cadc-rest/build.gradle
jburke-cadc: Removed.

cadc-log/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
jburke-cadc: Removed the class from the logging, duplicate information.

cadc-log/src/main/java/ca/nrc/cadc/log/ServletLogInfo.java
jburke-cadc: It's not. But runID is useful and should be set if available. Will check for it and set if available.

PR #108, opened 2019-11-21
S2524 - allow log level changes using a certificate

jburke-cadc:


cadc-util/src/main/java/ca/nrc/cadc/auth/AuthenticationUtil.java
```diff
@@ -272,9 +272,10 @@ public static AuthMethod getAuthMethodFromCredentials(Subject subject) {
      * </p>
      *
      * @param principalExtractor The PrincipalExtractor to provide Principals.
+     * @param augmentSubject Whether to augment the subject using Authenticator interface.
```
andamian: How are these changes related to `logControl`?

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -359,21 +390,122 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+
         if (authorizerClassName != null && groupAuthorizer == null) {
             throw new RuntimeException("CONFIG: group authorizer was configured but failed to load: " + authorizerClassName);
         }
 
+        Subject subject = AuthenticationUtil.getSubject(request);
+        logger.debug(subject.toString());
+
+        // Check group membership using config file properties
+        Set<String> groupUris = getAuthorizedGroupUris();
```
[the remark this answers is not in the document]
jburke-cadc: Updated to add the servlet init groupURI to the group list.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -359,21 +390,122 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+                    return;
+                } catch (AccessControlException ignore) {
+                    // failover to next auth method
+                }
+            }
+        }
+
+        // Check membership using servlet init parameters
         if (groupAuthorizer == null) {
             logger.warn("Authorization not configured, log control is public.");
```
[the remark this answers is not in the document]
jburke-cadc: Done.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -359,21 +390,122 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+            if (isAuthorizedUser(request)) {
+                return;
+            }
+        } catch (Exception e) {
+            if (e instanceof AccessControlException) {
+                logger.warn("User authorization failed: " + e.getMessage());
+            }
+        }
+
         if (authorizerClassName != null && groupAuthorizer == null) {
```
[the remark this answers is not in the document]
jburke-cadc: Removed.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -227,6 +255,9 @@ public void init(final ServletConfig config) throws ServletException {
             }
         }
 
+        // get the logControl properties file for this service if it exists
```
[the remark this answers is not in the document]
jburke-cadc: Removed.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -359,21 +366,147 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
-        if (authorizerClassName != null && groupAuthorizer == null) {
-            throw new RuntimeException("CONFIG: group authorizer was configured but failed to load: " + authorizerClassName);
+    private void authorize(HttpServletRequest request, boolean readOnly)
+        throws AccessControlException, TransientException {
+
+        // Track exceptions thrown
+        AccessControlException acException = null;
+
+        // first check if request user matches authorized config file users
+        Set<Principal> authorizedUsers = getAuthorizedUserPrincipals();
```
pdowler: The config file is read twice instead of once. It should be read once and used for the remainder of the request. You just need to re-org use of getLogControlProperties (called once, read once, Properties object passed to other methods).

cadc-log/src/test/java/ca/nrc/cadc/log/LogControlServletTest.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import javax.security.auth.x500.X500Principal;
+import javax.servlet.ServletConfig;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import org.apache.log4j.Level;
+import org.apache.log4j.Logger;
+import org.easymock.EasyMock;
```
pdowler: I hate easymock

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
jburke-cadc: Updated to read the properties file once per request.

cadc-log/src/test/java/ca/nrc/cadc/log/LogControlServletTest.java
jburke-cadc: Yeah too easy to do the mock just to pass the test.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -560,9 +562,8 @@ void isAuthorizedGroup(Authorizer authorizer, Subject subject, boolean readOnly)
      *
      * @return Set of authorized groupURI's, can be an empty Set if none configured.
      */
-    Set<String> getAuthorizedGroupUris() {
+    Set<String> getAuthorizedGroupUris(PropertiesReader propertiesReader) {
```
pdowler: This should return Set of GroupURI

jburke-cadc: Updated to use the GroupClient to validate group membership using the properties file, and return a list of GroupURI's when reading the groups in the properties file.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -359,29 +371,173 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+
+        // Get the logControl properties if they exist.
+        PropertiesReader propertiesReader = getLogControlProperties();
+
+        // first check if request user matches authorized config file users
+        Set<Principal> authorizedUsers = getAuthorizedUserPrincipals(propertiesReader);
+        if (authorizedUsers.isEmpty()) {
+            logger.debug("Authorized users not configured");
+        } else {
+            try {
```
pdowler: By throwing an exception as a response to what is essentially a question, this is using a try/catch to do logic. It would be cleaner for doUserCheck to return true if the caller is in the list and the code read something like:
`if (doUserCheck(caller, authorizedUsers)) {
   return;
}`
Below, once a doGroupCheck(caller, groupUris) returns true you can also return immediately.

Then try the Authorizer (which you have).

The last line of this method should be 
`throw new AccessControlException(...)`

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -359,29 +371,173 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
     /**
      * Check for proper group membership.
      */
-    private void authorize(HttpServletRequest request, boolean readOnly) throws AccessControlException, TransientException {
-        if (authorizerClassName != null && groupAuthorizer == null) {
-            throw new RuntimeException("CONFIG: group authorizer was configured but failed to load: " + authorizerClassName);
+    private void authorize(HttpServletRequest request, boolean readOnly)
+        throws AccessControlException, TransientException {
+
```
pdowler: The start of this method should create a Subject for the caller and pass that into other methods. That's something that has to happen once so should be here. You added an optional lazy augment to AuthenticationUtil -- use it here.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -359,29 +371,173 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+            if (acException != null) {
+                throw acException;
+            } else {
+                logger.info("Authorization not configured, log control is public.");
+                return;
+            }
         }
+        logger.debug("Configured GroupURI's: " + groupUris);
 
+        // Augment the calling subject.
```
pdowler: If you have a non-augmented subject from above, I'm not sure you need an augmented one here. You have a DN for the caller, which is enough to do a group check. 

I also do not see any call to cred to load a client proxy cert so I don't see how the calls to /ac are going to work. That has to happen once before looping over groups.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -359,21 +366,150 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+     *
+     * @param request The Http request.
+     * @param authorizedUsers The set of authorized user principals.
+     * @return true if the calling Principal matches an authorized principal.
+     */
+    boolean isAuthorizedUser(HttpServletRequest request, Set<Principal> authorizedUsers)
+        throws AccessControlException {
+        String callerName = "unknown";
+        ServletPrincipalExtractor principalExtractor = new ServletPrincipalExtractor(request);
+        Set<Principal> principals = principalExtractor.getPrincipals();
```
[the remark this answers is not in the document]
jburke-cadc: I don't see a ServletPrincipalExtractor method getPrincipals(X500Principal.class)?

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -395,21 +531,93 @@ private void authorize(HttpServletRequest request, boolean readOnly) throws Acce
...
+     * @return Set of authorized X500Principals, can be an empty Set if none configured.
+     */
+    Set<Principal> getAuthorizedUserPrincipals(PropertiesReader propertiesReader) {
+        Set<Principal> principals = new HashSet<Principal>();
+        if (propertiesReader != null) {
+            try {
+                List<String> properties = propertiesReader.getPropertyValues(USER_DNS_PROPERTY);
+                if (properties != null) {
+                    for (String property : properties) {
+                        if (StringUtil.hasLength(property)) {
```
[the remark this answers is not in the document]
jburke-cadc: The getPropertyValues can return a null if the property isn't in the properties file.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
jburke-cadc: Updated the auth checking methods to return a boolean and removed the try clauses.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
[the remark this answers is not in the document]
jburke-cadc: But if the method takes a Subject instead of the request, then that method can be used.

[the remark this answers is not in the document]
pdowler: Oh, this comment ended up out of place. Once you change to creating an un-augmented subject and passing that in, caller.getPrincipals(X500Principal.class) returns a correctly typed Set.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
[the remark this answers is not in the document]
pdowler: But that for loop won't ever have a null loop variable. That was the "property" variable I meant.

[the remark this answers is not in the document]
jburke-cadc: Sorry, looking at the wrong loop. Updated.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
pdowler: This is the comment that relates to getPrincipals(X500Principal.class) above

jburke-cadc: Changed to not augment the Subject.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
jburke-cadc: Added a call to cred to get a client proxy cert.

PR #110, opened 2020-01-07
new content expectation exception

brianmajor:


cadc-util/src/main/java/ca/nrc/cadc/net/ContentExpectationException.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+package ca.nrc.cadc.net;
+
+/**
+ * Exception to convey that the received content did not meet the
+ * checksum or length expectations provided by the client.
+ * 
+ * @author majorb
+ */
```
pdowler: In http Expect failures use code 417 and are linked to the Expect header... so this name isn't the best.
We could go more explicit (CorruptContentException) or even two very explicit like IncorrectContentChecksumException and IncorrectContentLengthException (since we don't need to report both)

Or we could do with something like the http name (precondition)... PreconditionFailedException? that ties it more tightly to HTTP 412 so I'm not a huge fan if it is in cadc-util.

PR #112, opened 2020-01-10
Augment subject when making LogControl requests

cbanek:
So in the LogControlServlet, when a post or a get happens, the auth
data and AuthenticatorImpl are used, but by passing in false to
getSubject, the subject is not augmented.

For LSST, we start off with a token credential, then when we augment
this we pull out data such as the username, userid, etc.  So to have
the username to check against, the subject must be augmented.

### DIFF
--- a/cadc-log/build.gradle
+++ b/cadc-log/build.gradle
@@ -15,7 +15,7 @@ sourceCompatibility = 1.7
 
 group = 'org.opencadc'
 
-version = '1.1.1'
+version = '1.1.2'
 
 dependencies {
     compile 'log4j:log4j:[1.2,)'
--- a/cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
+++ b/cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
@@ -373,8 +373,17 @@ public class LogControlServlet extends HttpServlet {
     private void authorize(HttpServletRequest request, boolean readOnly)
         throws AccessControlException, TransientException {
 
-        // Get the calling subject.
-        Subject subject = AuthenticationUtil.getSubject(request);
+        // Get the calling subject.  If possible, augment the subject,
+        // but if augmenting fails, use the subject provided only.
+        Subject subject;
+
+        try {
+            subject = AuthenticationUtil.getSubject(request);
+        } catch (Exception e) {
+            logger.error("Augment subject failed, using non-augmented subject: " + e.getMessage());
+            subject = AuthenticationUtil.getSubject(request, false);
+        }
+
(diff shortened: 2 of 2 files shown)


pdowler:
Although we kept the backwards compatible "config one group in the web.xml" we recently introduced the ability to have a config file (outside the war, in $HOME/config) that granted log control access to users (by X500 DN) and/or groups. The servlet now tries to do a lazy init and not augment so that it would be able to let ops staff change log levels even when the augment was failing for some reason (eg some or all of A&A was not working/reachable). 

I think the bit that is missing is the second half of the lazy augment (where the code does actually do an augment eventually and then try or retry authorization checks. Am I correct in assuming that you rely on a group membership check via the GroupClient (line ~420 for the backwards compat check) but it relies on the augment? Since we only use this for ops and they use an X509 proxy cert it happens that the CDP and GMS calls work with only the DN. 

So, the fix for the intended behaviour is to perform the augment again (without the false arg) before the try block at line ~400.

If you want to move to the external config with some direct user permissions then some additional work would be needed (isAuthorizedUser only supports granting to an X500Principal is the thign that works without augment). You could use the external file with group(s) with less work.

cbanek:
Right so for me, I am trying to use the file that has the list of groups and users that should be able to control the log endpoint.  Right now it's just users = CN=cbanek (which is me).  So I'm hoping to get in on the block in 383-388.  

Now, in order to get the username, I have to do the augment.  This takes the opaque token and puts all the principals in (including an X500Principal with my name, which comes from the token).  This is the X500Principal I'm checking against in 383-388.

With this PR and doing the augment right away, this all seems to work for me.

I was thinking that groups would actually be harder, because after looking at a bunch of things, like the GroupMembershipCheck and other interfaces, I think there's a fair amount of things I'm not sure I understand... like how do I put a group on a principal?  Maybe I've misunderstood the group checking.

I get what you mean about not doing the augment to help ops, since the augment might fail.  Although for LSST, the augment is pretty much always necessary, since the token doesn't give you anything.  What about maybe making a system property to see if you want to augment or not for this just as another idea.  I hate configuration so I understand this seems a bit hacky.

cbanek:
Or just one more idea, maybe if the augment fails, then just try to do it without the augment in a catch? (but keep going with whatever creds you have on hand)

cbanek:
Pushed a change where I fall back to using the non augmented subject.  I was trying to maybe just do a try with the augment, but it looks like that requires more code since many of those functions are private.  If this direction seems good but there are better functions to call let me know!

cbanek:
Thanks for circling back on this, totally fell off my radar.  Should I increment the version of the logging library?  I think I missed that.  not sure if that will be happened at some other level?

PR #114, opened 2020-01-16
Fixes #70

### LINKED ISSUE #70: HttpUpload: read error from response body
When HttpUpload receives an error response from the host, it should read the body of the response to gather any plain text error messages.  These should be assigned to the responseBody variable (as it is with a successful upload) so that clients of HttpUpload have access.

at88mph:
Obtain the proper error message from the Error Input Stream, if present, for the `HttpUpload` class.

### DIFF
--- a/cadc-util/build.gradle
+++ b/cadc-util/build.gradle
@@ -15,7 +15,7 @@ sourceCompatibility = 1.7
 
 group = 'org.opencadc'
 
-version = '1.2.29'
+version = '1.2.30'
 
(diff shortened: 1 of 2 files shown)


at88mph:
I see that now.  That `readResponse()` method from `HttpPost` _could_ be put into `HttpTransfer`, but it doesn't handle the case where the incoming `InputStream` is null, as can be the case with `getErrorStream()`.  Is it safe to reuse that one?

PR #118, opened 2020-02-20
refactor ca.nrc.cadc.net Http classes

pdowler:
- extend error handling to map http response code to exceptions (reverse of cadc-rest)
- add HttpTransfer.prepare() method that performs action (prepares the input stream) and throws
- run() still prepares, reads response, and captures exceptions


cadc-util/src/main/java/ca/nrc/cadc/net/HttpDownload.java
```diff
@@ -458,27 +388,26 @@ private void runX()
             log.debug("caught: " + tex);
             throwTE = true;
             throw tex;
-        } catch (AccessControlException ex) {
+        } catch (AccessControlException | ExpectationFailedException | IllegalArgumentException | PreconditionFailedException | ResourceNotFoundException ex) {
```
brianmajor: I'm sure there's a good reason for this set of 400 level errors to be caught here but I don't know what that is.

cadc-util/src/main/java/ca/nrc/cadc/net/ExpectationFailedException.java
```diff
@@ -65,117 +65,24 @@
...
-import ca.nrc.cadc.db.DBUtil;
-import ca.nrc.cadc.util.Log4jInit;
-import javax.sql.DataSource;
-import org.apache.log4j.Level;
 import org.apache.log4j.Logger;
-import org.junit.Assert;
-import org.junit.Test;
 
 /**
  *
```
brianmajor: I find it nice to have a brief description of what the exception means in the class level javadoc.

cadc-util/src/main/java/ca/nrc/cadc/net/HttpDownload.java
pdowler: removed these catch blocks

cadc-util/src/main/java/ca/nrc/cadc/net/ExpectationFailedException.java
pdowler: added one liner (cut and paste from http definition of 417)

pdowler:
Before merging, I am going to put some thought into which exceptions should be declared RuntimeException and which are (checked) Exception... I think anything that a client could know a priori will never happen because it doesn't do the thing that might trigger it should be unchecked (Runtime) so the client does not have to declare it in the throws. That was a client throws clause will actually be meaningful.

PR #123, opened 2020-03-30
support for cadc-rest InitAction

pdowler:
also improve PropertiesReader/MultiValuedProperties API


pdowler:
found issue with PropertiesReader... hang on before review

PR #126, opened 2020-05-29
fix to always capture error message from body, truncate if too large

pdowler:


cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
```diff
@@ -1195,8 +1181,13 @@ private String readResponseBody(HttpURLConnection conn)
     
     protected String readResponseBody(InputStream istream)
             throws IOException, InterruptedException {
+        ByteCountInputStream bcis = new ByteCountInputStream(istream, maxReadFully);
         try (ByteArrayOutputStream byteArrayOstream = new ByteArrayOutputStream()) {
```
brianmajor, line 1185: This is a new form of syntax to me.  Looks appropriate.

PR #128, opened 2020-06-16
implemented MultiBufferIO with async write thread; deprecated ThreadedIO

pdowler:


cadc-util/src/main/java/ca/nrc/cadc/io/DiscardOutputStream.java
```diff
@@ -0,0 +1,99 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import org.apache.log4j.Logger;
+
+/**
+ * OutputStream that discards all bytes.
+ * 
+ * @author pdowler
+ */
+public class DiscardOutputStream extends OutputStream {
```
brianmajor, line 79: Is this in the main src tree because it could be used by tests outside of cadc-util?

pdowler: Yes, I also need this in some storage adapter tests, and something like it was needed in a critwall test

PR #129, opened 2020-09-17
improve RsaSignature class usage

pdowler:
also an error handling change in cadc-rest that is kind of old (was in my fork) but no library version increment; can't recall what motivated that change


cadc-rest/src/main/java/ca/nrc/cadc/rest/RestAction.java
```diff
@@ -360,8 +355,11 @@ protected void handleException(
                     cause = cause.getCause();
                 }
             }
-
-            os.write(sb.toString().getBytes());
+            byte[] out = sb.toString().getBytes();
+            syncOutput.setHeader("Content-Length", Integer.toString(out.length));
+            
+            OutputStream os = syncOutput.getOutputStream();
+            os.write(out);
```
brianmajor, line 362: For some reason in /ac running in proto as a container I've had to put an out.flush() after the write, otherwise nothing comes out.

brianmajor (approved):

PR #130, opened 2020-10-02
s2803 -  SyncOutput: open the OutputStream when data is written.

jburke-cadc:


pdowler (changes requested):
I would comment out or remove the debug message in the `write` methods. Those get called  millions of times in very tight loops when streaming large files... even with a constant string it could be bad.

jburke-cadc:
The debug line in the write methods should log only once when the OutputStream is first opened. Moved the debug into the if (out == null) loop after overly verbose logging when it's outside the loop.

pdowler (approved):

PR #133, opened 2020-11-05
CADC-50 auth token support

brianmajor:


brianmajor:
Will restart with a new pull request

PR #135, opened 2020-12-02
CADC-9705 CADC-9706 Updated WebServiceLogInfo to handle attributes of type List<String>

yeunga:


cadc-log/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
```diff
@@ -165,6 +166,29 @@ private String getPreamble() {
         return sb.toString();
     }
 
+    // handle Lists, refer to GroupLogInfo.addedMembers
+    // currently only List<String> is handled
+    private void populateList(StringBuilder sb, Object o) {
```
brianmajor, line 171: Why not send in the List<?> rather than the object?  Then you won't need to re-cast.

cadc-log/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
```diff
@@ -165,6 +166,29 @@ private String getPreamble() {
...
+        sb.append("[");
```
brianmajor, line 172: Were you able to test that the square brackets work with the elasticsearch/kibana log scraping?

brianmajor (changes requested):

cadc-log/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
yeunga: Changed.

cadc-log/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
```diff
@@ -165,7 +166,31 @@ private String getPreamble() {
...
     }
 
-    private void populate(StringBuilder sb, Class c) {
+    // handle Lists, refer to GroupLogInfo.addedMembers
+    // currently only List<String> is handled
+    private void populateList(StringBuilder sb, List<?> o) {
+        sb.append("[");
+        if (!(o.isEmpty())) {
+            boolean isFirstElement = true;
+            if (o.get(0) instanceof String) {
```
brianmajor, line 175: If this is not an instance of List<String> then the value is lost.  It should have the same behaviour as before if not an instance of List<String>.

brianmajor (changes requested):

cadc-log/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
yeunga: o has to be an instance of List at this point.

cadc-log/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
yeunga: No, but this is in the JSON spec. Elasticsearch should handle it. I suppose we can check kabana log to see if it handles an empty list. If there an easy to test this out on Elasticsearch?

yeunga: https://www.elastic.co/guide/en/elasticsearch/reference/current/null-value.html, and https://stackoverflow.com/questions/43326775/elastic-search-query-empty-list
indicates that ElasticSearch handles an empty list []

cadc-log/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
brianmajor: I mean the case where o is, for example, an instance of List<Integer>.  Why not just check for List<String> instead of List<?> in populate()?

brianmajor (changes requested):

cadc-log/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
```diff
@@ -172,18 +174,17 @@ private void populateList(StringBuilder sb, List<?> o) {
...
-                    sb.append(element);
-                    sb.append("\"");
+            for (Object element : o) {
+                String val = sanitize(element);
+                if (isFirstElement) {
+                    isFirstElement = false;
+                } else {
+                    sb.append(",");
+                }
+                if (element instanceof String || element instanceof UUID) {
```
brianmajor, line 184: Looks much better.  I guess the only question left is whether to put double quotes around all list elements (not just String and UUID).  This is a question of whether `o` is a complex object or a primitive.  Since it's going through the sanitize() method already, I think we can only support non-complex elements.  So, what do you think of removing this check (line 184) and putting double quotes around all elements?

brianmajor (changes requested):

cadc-log/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
yeunga: If we do that, List<Integer> would produce incorrect JSON format, e.g. "intList":["1","2","3"]. With the simple-minded algorithm that we have currently implemented, we cannot expect to be able to cover all cases. We should expect developers who will add new attributes in the future to update the algorithm to produce the correct JSON for log info.

yeunga: I added UUID in the unit test to show that even for simple classes, toString() does not solve the problem for us. In the case of UUID, we cannot just enclose the sanitized value within [ and ] but need to add " and " instead.

brianmajor: Okay, sounds good.

PR #137, opened 2020-12-17
CADC-8954: filter http response codes < 100 correctly in HttpTransfer.checkErrors()

hjeeves:
The check for '-1' would never be reached because the filter only worked for >= 400 errors. 

Part for CADC-1289: make storage UI work against cavern.

### DIFF
--- a/cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
+++ b/cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
@@ -760,7 +760,7 @@ public abstract class HttpTransfer implements Runnable {
         
         captureResponseHeaders(conn);
         
-        if (responseCode <= 100 && responseCode < 400) {
+        if (100 <= responseCode && responseCode < 400) {
             return;
         }
         log.debug("error: " + contentType + " " + contentLength);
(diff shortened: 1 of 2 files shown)


cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
```diff
@@ -760,7 +760,7 @@ protected void checkErrors(URL url, HttpURLConnection conn)
         
         captureResponseHeaders(conn);
         
-        if (responseCode < 400) {
+        if (responseCode <= 100 && responseCode < 400) {
```
pdowler, line 763: In slack I wrote this in the slightly odd way (A <= B && B < C) because I can read it like math: A < B < C.

Here, you wrote it with the variable on the left of the operator instead of the right so it isn't "between" any more.

pdowler (changes requested):

cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
hjeeves: ok.. hoping this is right now.

pdowler (approved):

PR #143, opened 2021-02-23
CADC-8679 - added Digest header support to HttpTransfer

jburke-cadc:


cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
```diff
@@ -429,6 +460,110 @@ public String getContentMD5() {
...
+                                                                 + "<algorithm>=<checksum value> found %s",
+                                                             digest));
+        }
+
+        Algorithm algorithm;
+        try {
+            algorithm = Algorithm.fromString(scheme);
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException(String.format("Unsupported algorithm %s, supported algorithm's: "
+                                                                 + "md5, sha-1, sha-256, sha-384, sha-512", scheme));
```
brianmajor, line 506: Just a minor point: this list of algorithms will need to be maintained to match the enum, so perhaps remove the list or have it come from the enum.

brianmajor (changes requested):
Looks good.  Just a few minor things and a question:  will minoc need some of the parsing mechanisms in here to set the DIGEST via SyncOutput?  If so perhaps it could be broken out into it's own class or into an existing utility class?  I see that it's public and static right now, HttpTransfer isn't really used that way currently.

jburke-cadc:
Pat and I discussed this a bit. One comment was this was temporary. But rereading that was for modifying the Content-MD5 parsing, not simply adding Digest support. Splitting the digest code into a separate class is be a good idea, which I will do.

cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
```diff
@@ -429,6 +460,110 @@ public String getContentMD5() {
     public Date getLastModified() {
         return lastModified;
     }
+
+    /**
+     * Disgest checksum from http header.
```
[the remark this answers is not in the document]
jburke-cadc: Updated

cadc-util/src/test/java/ca/nrc/cadc/date/DateUtilTest.java
```diff
@@ -245,7 +246,9 @@ public void testISO8601()
         }
     }
 
+    @Ignore
     @Test
+    // Broken and could not find any useage of HTTP_DATE_FORMAT
```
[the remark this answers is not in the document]
jburke-cadc: Updated.

cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
jburke-cadc: Yes, the list of algorithms should not be hard coded in the exception message. Will skip the list, or get from the enum.

cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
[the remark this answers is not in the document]
brianmajor: Hmm, digest still spelt wrong

cadc-util/src/test/java/ca/nrc/cadc/net/HttpTransferTest.java
```diff
@@ -70,13 +70,17 @@
...
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
 import java.util.Date;
```
brianmajor, line 84: These imports can be removed... not a big deal.  I guess there's no checksytle on test classes.

cadc-util/src/main/java/ca/nrc/cadc/net/Digest.java
```diff
@@ -0,0 +1,225 @@
...
+import ca.nrc.cadc.util.StringUtil;
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import org.apache.log4j.Logger;
+
+/**
+ * Class to represent a Digest HTTP header.
+ */
+public class Digest {
```
brianmajor, line 81: Maybe DigestUtil instead of Digest?  It's currently a class with no state/data.  We could either:
- Keep it as Digest.java and add a constructor that took the string and did the parsing right away
- Rename it to DigestUtil.java and keep it as is.
Thoughts?

jburke-cadc: It doesn't represent a Digest, so I think rename to DigestUtil would be best. The static methods will likely be the only use of the class.

cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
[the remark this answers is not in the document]
jburke-cadc: Fixed, again.

cadc-util/src/test/java/ca/nrc/cadc/net/HttpTransferTest.java
jburke-cadc: Updated the spelling and removed unused imports. No, checkstyle doesn't run on test classes, not sure why the distinction.

brianmajor (approved):

cadc-util/src/main/java/ca/nrc/cadc/net/DigestUtil.java
```diff
@@ -0,0 +1,225 @@
...
+                if (algorithm.value.equalsIgnoreCase(value)) {
+                    return algorithm;
+                }
+            }
+            throw new IllegalArgumentException(String.format("Algorithm not found for value %s", value));
+        }
+    }
+
+    /**
+     * Parse the given digest string and return a URI of the form: {algorithm}:{checksum in hex}
```
pdowler, line 111: The above description is not correct.

cadc-util/src/main/java/ca/nrc/cadc/net/DigestUtil.java
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        Algorithm algorithm;
+        try {
+            algorithm = Algorithm.fromString(scheme);
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException(String.format("Unsupported algorithm %s", scheme));
+        }
+
+        // For a MD5 checksum, determine if the checksum value is hex or base64
```
pdowler, line 146: Noooooo!!!!! Don't re=introduce the old bug here! Spec says md5 values are base64 encoded so only support that.

pdowler (changes requested):

pdowler (approved):

cadc-util/src/main/java/ca/nrc/cadc/net/DigestUtil.java
jburke-cadc: Fixed the uri description.

PR #147, opened 2021-04-14
CADC-9182 Added checkReadable() and checkWritable() as suggested by P…

yeunga:
…at in his code review.

### DIFF
--- a/cadc-rest/src/main/java/ca/nrc/cadc/rest/RestAction.java
+++ b/cadc-rest/src/main/java/ca/nrc/cadc/rest/RestAction.java
@@ -195,7 +195,9 @@ public abstract class RestAction implements PrivilegedExceptionAction<Object> {
 
 
     /**
-     * Check if the caller can read the resource.
+     * Check the service state to determine if a read should go ahead.
+     * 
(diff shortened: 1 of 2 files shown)


cadc-rest/src/main/java/ca/nrc/cadc/rest/RestAction.java
```diff
@@ -193,6 +193,28 @@ protected void initState() {
         }
     }
 
+
+    /**
+     * Check if the caller can read the resource.
```
pdowler, line 198: This is not a permission check so has nothing to do with the specific caller. This is checking the service state to see if any read should go ahead. Please clarify the javadocs and also add an `@throws IllegalStateException` (it's a RuntimeException, so not requried but good documentation for implementers of subclasses.

also: the checkWritable javadocs
otherwise, code is fine

pdowler (changes requested):

cadc-rest/src/main/java/ca/nrc/cadc/rest/RestAction.java
yeunga: Updated.

pdowler (approved):

PR #150, opened 2021-04-23
OAuth tokens, auth library reorg - CADC-50

brianmajor:


brianmajor:
Ready for another look @pdowler

cadc-util/src/main/java/ca/nrc/cadc/auth/Authenticator.java
```diff
@@ -72,22 +72,39 @@
...
  *
  * @author pdowler
  */
 public interface Authenticator {
+    
+    /**
+     * If necessary, validate any principals in the subject.  Some principals,
+     * such as X500Principal, do not require validation as that is done with
+     * TLS.  Others, such as cookies and tokens, need to be validated.
+     * Failed validation must result in an AccessControlException
```
pdowler, line 85: javadoc and throws inconsistent

The expected usage needs to be documented and this seems like the place. My understanding is that the incoming subject could contain an **unvalidated** AuthorizationTokenPrincipal and a typical validate would do something  like:
* parse/validate the AuthorizationTokenPrincipal and put the resulting AuthorizationToken into the pub credentials
* replace the AuthorizationTokenPrincipal with an HttpPrincipal if a username can be extracted, or wait for augment(...), or do do nothing... but should probably say that an authenticated user must have a principal so just removing it would be bad

That would be the same for a CookiePrincipal. 
Some related comments below in AuthorizationTokenPrincipal and AuthorizationToken

cadc-util/src/main/java/ca/nrc/cadc/auth/AuthorizationTokenPrincipal.java
```diff
@@ -0,0 +1,113 @@
...
+ *                                       <http://www.gnu.org/licenses/>.
+ *
+ ************************************************************************
+ */
+
+package ca.nrc.cadc.auth;
+
+import java.security.Principal;
+
+/**
```
pdowler, line 72: This is the unvalidated "incoming" token that Authenticator.validate should examine. Should state that usage and the relationship to AuthorizationToken.

Is the `token` value here the raw header value? And the validate() method parses it into an AuthorizationToken?

Why does this class track the http header that was used? Just backwards compat for DelegationToken? If so, then DelegationToken should extend this and the 2-arg ctor should be protected.

cadc-util/src/main/java/ca/nrc/cadc/auth/AuthorizationToken.java
```diff
@@ -0,0 +1,117 @@
...
+ *
+ ************************************************************************
+ */
+
+package ca.nrc.cadc.auth;
+
+/**
+ * Class representing an authorization token.  For example, an opaque OAuth2 token value
+ * with name 'Bearer'.
+ *
```
pdowler, line 73: This is the object that Authenticator.validate would put into the subject's credential(s)? Should say that this is the validated token usable as credential.

Why doesn't it also keep track of the http header to use? Is it always `Authorization`?

cadc-util/src/main/java/ca/nrc/cadc/auth/DelegationToken.java
```diff
@@ -126,7 +125,7 @@
 
     public static final String FIELD_DELIM = "&";
     public static final String VALUE_DELIM = "=";
-
+
```
pdowler, line 128: It looks like DelegationToken is losing all scope support and scope validation. Why not just remove it if it is no longer supported? 

I know it was re-used with an "sso" scope for cookies and login tokens, but then it should be renamed to something that says what it's for: PKI based token generate and validate. "Delegation" was meant to convey that a party other than the user and the service could acquire one of these (safely because of scope).

Is the delegation and scope part no longer used/supported in vault? There is still a VOSScopeValidator.java in there

cadc-util/src/main/java/ca/nrc/cadc/auth/BearerTokenPrincipal.java
```diff
@@ -75,12 +75,17 @@
 import java.security.Principal;
 
 /**
- * Represents a bearer token
+ * Represents a bearer token.
+ * 
+ * @deprecated Being replaced with more general token support in AuthorizationToken
+ *     and AuthorizationTokenPrincipal
  */
+@Deprecated
```
pdowler, line 83: If in addition to deprecating BearerTokenPrincipal you made it a subclass of AuthorizationToken you wouldn't need to add two principals for compatibility. You could just add this thing for awhile and then switch to AuthorizationToken when this class gets deleted. That's a common and useful pattern when a class gets replaced by something else.

cadc-util/src/main/java/ca/nrc/cadc/auth/ServletPrincipalExtractor.java
```diff
@@ -138,64 +135,44 @@ public ServletPrincipalExtractor(final HttpServletRequest req) {
...
 
+        // authorization header (bearer and token)
         String authToken = request.getHeader(AuthenticationUtil.AUTHORIZATION_HEADER);
         if (BearerTokenPrincipal.isBearerToken(authToken)) {
-            principals.add(new BearerTokenPrincipal(authToken));
+            // deprecated in favour of a common token handling mechanism
+            BearerTokenPrincipal bearerTokenPrincipal = new BearerTokenPrincipal(authToken);
+            principals.add(bearerTokenPrincipal);
+        }
+
```
pdowler, line 152: see comment in BearerTokenPrincipla about having to add two principals...

pdowler (commented):
Review comments for changes in cadc-util

cadc-util/src/main/java/ca/nrc/cadc/auth/AuthorizationTokenPrincipal.java
brianmajor: Yes, just for backwards compatibility.  I'll try doing the subclass trick instead.

cadc-util/src/main/java/ca/nrc/cadc/auth/AuthorizationToken.java
brianmajor: Yes, this is always from the `Authorization` header.

cadc-util/src/main/java/ca/nrc/cadc/auth/DelegationToken.java
brianmajor: For vault, the scope validator compared the request path with the node, so it was scoped on to a node.  During this refactoring we found that getting access to the servlet HTTP request object at the time of scope validation was difficult.  Since the functionality wasn't being used we decided to drop scope support and re-introduce it later when needed.

I think I tried to remove ScopeValidator altogether and found some complications somewhere.  I'll give it another shot.  And I'll rename DelegationToken to, say, SignedToken?

brianmajor: I've left the scope field in DelegationToken because I think it will need to go there when we revisit scope again.  I'm using it in the OAuth/OpenID work too.  But, the validation parts are gone, so clients need to do validation themselves if using it.

brianmajor: DelegationToken.java (now SignedToken.java) is a bit of a mess and would be difficult to fit into a hierarchy with AuthenticationToken.  (The contructors don't follow the same pattern.). Not sure I want to take on the cleanup as part of this pull request as there are already many changes.

cadc-util/src/main/java/ca/nrc/cadc/auth/Authenticator.java
pdowler: typo in the javadoc "AuthoricationToken"

cadc-util/src/main/java/ca/nrc/cadc/auth/DelegationToken.java
pdowler: DelegationToken (now SignedToken) is more a utility to create AuthorizationToken(s) That can be validated with the pub key (instead of a database lookup or calling back to the login service, for example). Over in cadc-permissions there was something similar to sign a tuple with {user,grant,asset uri} and it was confusing until I turned it into TokenTool with methods to create and validate a specific tuple... so I don't think it needs to go into the set of token objects and we can worry about cleaning up some more later.

cadc-util/src/main/java/ca/nrc/cadc/auth/TokenValidator.java
```diff
@@ -0,0 +1,158 @@
...
+                if (!AuthenticationUtil.CHALLENGE_TYPE_BEARER.equals(challengeType)
+                    && !AuthenticationUtil.CHALLENGE_TYPE_IVOA.equals(challengeType)) {
+                    throw new NotAuthenticatedException(challengeType, OAuthError.INVALID_REQUEST,
+                        "unsupported challenge type: " + challengeType);
+                }
+            }
+            
+            try {
+                SignedToken validatedToken = SignedToken.parse(credentials);
+                subject.getPrincipals().add(validatedToken.getUser());
```
pdowler, line 144: I think you need to remove the AuthorizationTokenPrincipal from the subject. It's only a transient and no longer needed.

cadc-util/src/main/java/ca/nrc/cadc/auth/NotAuthenticatedException.java
```diff
@@ -78,12 +78,64 @@
  */
 public class NotAuthenticatedException extends SecurityException {
     
+    private String challenge;
+    private OAuthError oauthError;
+
```
pdowler, line 83: Aside from the fact that VO people will quibble about the actual values, I think this could be usable with ivoa challenge failures as well... as noted in RestServlet we should report failed authentication for ivoa challenge. And we will have expired ivoa tokens :-)

In theory, this could occur with client cert if the Authenticator.validate(...) impl imposed further restrictions not enforced by the CA bundle.

Suggestion:  maybe rename the things in code (enum, arg names) to just AuthError and have a comment that the words are from the OAuth2 API... just cosmetic really: the important thing is that our validate adds the OAuthError when throwing the exception with an ivoa challenge.

pdowler (changes requested):

cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java
```diff
@@ -296,8 +306,15 @@ protected void doit(HttpServletRequest request, HttpServletResponse response,  C
             action.setSyncInput(in);
             action.setSyncOutput(out);
             action.setLogInfo(logInfo);
+            
+            setAuthenticateHeaders(subject, out);
 
             doit(subject, action);
+        } catch (NotAuthenticatedException ex) {
+            logInfo.setSuccess(true);
+            logInfo.setMessage(ex.getMessage());
```
pdowler, line 315: The `setFailedAuthenticationHeader` method looks like it only handles OAuth2 errors (from the `getOauth2Error() = null` bit). I would try to include an error attr in an ivoa challenge in the same style; I'm sure that'll be acceptable to others.

Shouldn't all challenges be re-iterated when authentication fails? Might have to move the bulk of  `setAuthenticateHeaders(Subject,SyncOutput)` into a method that can be used in both scenarios so you can blend in the error message with the challenge that was attempted and still include the others. Maybe just `setAuthenticateHeaders(SyncOutput,Subject,NotAuthenticatedException)` and it's called with a null arg either way?

pdowler (changes requested):

cadc-util/src/main/java/ca/nrc/cadc/auth/NotAuthenticatedException.java
brianmajor: Yes, I think those names are misleading, I'll remove the OAuth.  I think (at least I intended) to make that error handling work with both types of tokens.  If you look in TokenValidator where the exceptions are constructed it doesn't differentiate.

cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java
brianmajor: Mentioned in the comment below but will again: it should work with both types of tokens and others we add.  The validator which throws them doesn't care what type they are.

Regarding setting all challenges even on failure:  hmm, not sure about that.  I thought it would be confusing because you'd have to find the header with the error message to diagnose your current problem.  Maybe we should discuss more tomorrow because I think it could go either way.

pdowler (approved):

PR #153, opened 2021-05-11
Added authMethod to WebServiceLogInfo

andamian:


cadc-util/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
```diff
@@ -253,6 +254,10 @@ public void setSuccess(boolean success) {
      */
     public void setSubject(Subject subject) {
         this.user = getUser(subject);
+        AuthMethod aMethod = AuthenticationUtil.getAuthMethod(subject);
+        if (aMethod != null) {
+            this.authMethod = aMethod.getValue();
```
pdowler, line 259: The authMethod could sometimes end up being null and other times "anon" depending on what code gets executed up to this point. Might be good to log consistently here: default to AuthMethod.ANON would be best.

pdowler (changes requested):

PR #161, opened 2021-07-07
Added DateUtil functions to assist with FITS conversions.

at88mph:
Small Javadoc cleanup.

### DIFF
--- a/cadc-util/src/main/java/ca/nrc/cadc/date/DateUtil.java
+++ b/cadc-util/src/main/java/ca/nrc/cadc/date/DateUtil.java
@@ -72,9 +72,10 @@ package ca.nrc.cadc.date;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Date;
-import java.util.HashMap;
+import java.util.List;
 import java.util.Locale;
 import java.util.TimeZone;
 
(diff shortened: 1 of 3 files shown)


cadc-util/src/main/java/ca/nrc/cadc/date/DateUtil.java
```diff
@@ -328,10 +352,81 @@ public static Date toDate(final long dmfSeconds) {
         return new Date((dmfSeconds * 1000L) + getDMFEpoch().getTime());
     }
 
+    /**
+     * Convert from the given MJD value to a Java Date object.
+     *
+     * @param modifiedJulianDate    The MJD value.
+     * @return  Date instance.  Never null.
+     */
+    public static Date fromModifiedJulianDate(final double modifiedJulianDate) {
```
pdowler, line 361: This is a more complicated method that does the same thing as `Date toDate(double)`, which is already known to be correct (using `slaDjcl`). I appreciate that the existing method name is pretty vague (relying on the name of the arg to have any meaning, so it would be fine to deprecate it and add the more explicitly `fromModifiedJulianDate`. The deprecated one should call the new one and the new one should have the code from the existing one.

Can you also deprecate the `Date toDate(long)` and `Date getDMFEpoch()` methods?  I think "dmf" was an HST-specific thing that is not used in any of our s/w. Will remove in the future.

pdowler (changes requested):

pdowler (approved):

PR #164, opened 2021-07-27
CADC-8668 added doShutdown() to InitAction and call in RestServlet destroy()

jburke-cadc:


cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java
```diff
@@ -160,22 +161,30 @@ public void init(ServletConfig config) throws ServletException {
...
         } catch (Exception ex) {
             log.error("init failed", ex);
             throw new ServletException("init failed", ex);
         }
     }
 
+    @Override
+    public void destroy() {
+        try {
+            initAction.doShutdown();
```
pdowler, line 182: `if (initAction != null)` ??

pdowler (changes requested):

cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java
jburke-cadc: Yes checking the initAction isn't null would stop a large unnecessary log message. Fixed.

pdowler (approved):

PR #172, opened 2021-11-10
Auth header changes

brianmajor:
note: the new `setAuthHeaders` boolean is there to prevent the infinite lookup loop that happens in the 2 credential issuing services:  ac and cred.

### DIFF
--- a/cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java
+++ b/cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java
@@ -487,7 +487,8 @@ public class RestServlet extends HttpServlet {
                 URI loginServiceURI = getLocalServiceURI(Standards.SECURITY_METHOD_PASSWORD);
                 URL loginURL = regClient.getServiceURL(loginServiceURI, Standards.SECURITY_METHOD_PASSWORD, AuthMethod.ANON);
                 StringBuilder sb = new StringBuilder();
-                sb.append(AuthenticationUtil.CHALLENGE_TYPE_IVOA_BEARER + " standard_id=\"").append(Standards.SECURITY_METHOD_PASSWORD.toString()).append("\", ");
+                sb.append(AuthenticationUtil.CHALLENGE_TYPE_IVOA_BEARER + " standard_id=\"")
+                    .append(Standards.SECURITY_METHOD_PASSWORD.toString()).append("\", ");
                 sb.append("access_url=\"").append(loginURL).append("\"");
                 appendAuthenticateErrorInfo(AuthenticationUtil.CHALLENGE_TYPE_IVOA_BEARER, sb, ex, false);
                 out.addHeader(AuthenticationUtil.AUTHENTICATE_HEADER, sb.toString());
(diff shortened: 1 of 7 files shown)


cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java
```diff
@@ -449,34 +463,51 @@ void setAuthenticateHeaders(Subject subject, SyncOutput out, NotAuthenticatedExc
...
+            log.debug("Setting " + AuthenticationUtil.VO_AUTHENTICATED_HEADER + " header");
+            Set<HttpPrincipal> useridPrincipals = subject.getPrincipals(HttpPrincipal.class);
+            if (!useridPrincipals.isEmpty()) {
+                HttpPrincipal userid = useridPrincipals.iterator().next();
+                out.addHeader(AuthenticationUtil.VO_AUTHENTICATED_HEADER, userid.getName());
+                return;
+            }
+            Set<Principal> allPrincipals = subject.getPrincipals();
+            if (!allPrincipals.isEmpty()) {
+                Principal principal = allPrincipals.iterator().next();
```
pdowler, line 478: First, you don't need to check isEmpty() again here and could go straight to 
`subject.getPrincipals().iterator().next();`

Seems like this will ~randomly pick an identity type when HttpPrincipal aka userID is not known... practically that's going to be an OpenIdPrincipal, an X500Principal, or maybe a PosixPrincipal ... It could even be a CookiePrincipal or AuthorizationTokenPrincipal if the AuthentricatorImpl.validate(Subject) didn't do anything. Mostly cosmetic, but maybe put a comment like `// TODO: pick the most informative principal`

cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java
```diff
@@ -449,34 +463,51 @@ void setAuthenticateHeaders(Subject subject, SyncOutput out, NotAuthenticatedExc
...
             RegistryClient regClient = getRegistryClient();
             
-            // set a header for info on how to obtain tokens with username/password over tls
+            // set a header for info on how to obtain bearer tokens with username/password over tls
             try {
                 URI loginServiceURI = getLocalServiceURI(Standards.SECURITY_METHOD_PASSWORD);
                 URL loginURL = regClient.getServiceURL(loginServiceURI, Standards.SECURITY_METHOD_PASSWORD, AuthMethod.ANON);
                 StringBuilder sb = new StringBuilder();
-                sb.append(AuthenticationUtil.CHALLENGE_TYPE_IVOA + " standard_id=\"").append(Standards.SECURITY_METHOD_PASSWORD.toString()).append("\", ");
+                sb.append(AuthenticationUtil.CHALLENGE_TYPE_IVOA_BEARER + " standard_id=\"")
```
pdowler, line 494: Mixing string + with append here (and below, ~508, ~527)

cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java
```diff
@@ -488,29 +519,24 @@ void setAuthenticateHeaders(Subject subject, SyncOutput out, NotAuthenticatedExc
                 log.debug("LocalAuthority -- not found: " + Standards.SECURITY_METHOD_OAUTH, notSupported);
             }
             
-            // TODO: mechanism to enable this rather than always set it
+            try {
+                // header for delegated x509 client proxy certificates
+                URI cdpProxyServiceURI = getLocalServiceURI(Standards.CRED_PROXY_10);
+                URL cdpProxyURL = regClient.getServiceURL(cdpProxyServiceURI, Standards.CRED_PROXY_10, AuthMethod.PASSWORD);
```
pdowler, line 525: Nice: working this out took care of that TODO to find a mechanism :-)

If you didn't use the shortcut `regClient.getServiceURL` this could be written look for CRED_PROXY_10 with either #BasicAA or #tls-with-password securityMethod instead of just basic... then if we changed /cred to use #tls-with-password, or someone else deployed differently, this would still work as expected. It would be something like:

```
Capabilities caps = regClient.getCapabilities(cdpProxyServiceURI);
Capability c = caps.findCapability(Standards.CRED_PROXY_10);
for (Interface i : c.getInterfaces()) {
    // find interface for suitable securityMethod to use as standard_id in challenge
}
```
The fact that this would work validates our approach because in principle we don't know what /cred provides but can extract it from caps and put it into challenge anyway.

pdowler (changes requested):

pdowler (approved):

PR #174, opened 2021-11-19
give rest actions initialize opportunity

brianmajor:


pdowler (changes requested):
I'm finding the duplicate catch(NotAuthenticatedException) blocks in RestServlet and RestAction are bugging me. If the one in RestAction was broken somehow, the one in RestServlet would still probably come into play, although not intentionally, and hide the fact that RestAction wasn't acting as expected.

It would be cleaner and more obvious if there was a 
```
try { ... } catch(NotAuthenticatedException) { ... }
catch(Throwable) { ... }
```
for just the AuthenticationUtil.getSubject (validate or augment failure) and then the existing try/catch  for creating and running the action (where all exceptions that are bugs). Then any exception that RestAction doesn't handle falls through to the "unexpected" and we'll detect the bug.

Thoughts?

brianmajor:
Yes, I agree with your suggestion.  I have done the rework.  Initially I put a try/catch (catching both NotAuthenticatedException and Throwable) in its own block, but found that the logging would have to be duplicated.  So, you'll see now that there's a try/catch around AuthenticationUtil.getSubject(), but only for NotAuthenticatedException.  This way logging and bug catching logic is in one place.  You'll see that it cleaned up the code in a few places too.

pdowler (approved):

PR #186, opened 2022-04-01
Expose server info to set Server response header.

at88mph:


pdowler (changes requested):
I don't think this satisfies the spirit of the IVOA language on this. The intent is for someone to figure out that "tap services based on OpenCADC libraries" always fail in this way, not "tomcat" or "java".

So, I think a better approach here would be for cadc-rest to simply add a default `server` header saying something like "OpenCADC/cadc-rest". Then we'd want libraries that implement actions to be able to provide a string that gets added, eg "OpenCADC/cadc-rest + cadc-uws-server". 

that would imply a `public String getServerImpl()` with a default that returns `null`. Then we can gradually add the detail(s) in other libs.

Also, it's considered bad/dangerous to output tomcat details and versions anyway (by SSC, who was after us on that topic awhile back).

pdowler (changes requested):
I guess I wasn't clear. The base RestAction class should actually set the `server` header as early as possible (line between line~337 and 338) rather than leaving that to (many) subclasses to handle.

So `getServerInfo()` would be private

at88mph:
I see now, thanks.  Fixed.

pdowler (approved):

PR #194, opened 2022-07-13
log value of x-forwarded-for

brianmajor:


cadc-util/src/main/java/ca/nrc/cadc/net/NetUtil.java
```diff
@@ -88,14 +88,17 @@
...
 
 /**
  * Miscellaneous network utility methods (static).
  *
  * @author pdowler
- * @version $Version$
  */
 public class NetUtil {
+    
+    private static Logger log = Logger.getLogger(HttpGet.class);
```
pdowler, line 101: this logger is created with wrong class

pdowler (changes requested):

PR #195, opened 2022-07-21
added resource and grant attrs to WebServiceLogInfo

pdowler:


cadc-util/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
```diff
@@ -211,6 +215,9 @@ private void populate(StringBuilder sb, Class<?> c) {
                         } 
                         sb.append("\"").append(f.getName()).append("\"");
                         sb.append(":");
+                        if (o instanceof URI) {
+                            o = ((URI) o).toASCIIString();
+                        }
                         if (o instanceof List<?>) {
```
brianmajor, line 221: `else if` to match the pattern?

brianmajor (changes requested):

cadc-util/src/main/java/ca/nrc/cadc/log/WebServiceLogInfo.java
pdowler: no, it's turning the value into a String so the next if/else will output as string (with quotes), so it has to be separate. If there were 2+ value types to convert, I would have put a method to make it more clear what's happening.

I'll add a comment.

PR #200, opened 2022-10-06
added AuthChallenge to parse/serialise www-authenticate challenges

pdowler:
added support for char[] param value in HttpPost

### DIFF
--- a/cadc-util/build.gradle
+++ b/cadc-util/build.gradle
@@ -16,7 +16,7 @@ sourceCompatibility = 1.8
 
 group = 'org.opencadc'
 
-version = '1.7.2'
+version = '1.7.3'
 
 description = 'OpenCADC core utility library'
 def git_url = 'https://github.com/opencadc/core'
--- a/cadc-util/build.gradle
+++ b/cadc-util/build.gradle
@@ -22,9 +22,16 @@ description = 'OpenCADC core utility library'
 def git_url = 'https://github.com/opencadc/core'
 
 dependencies {
-    compile 'org.apache.logging.log4j:log4j-1.2-api:[2.17.2,3.0)'
-    compile 'org.apache.logging.log4j:log4j-api:[2.17.2,3.0)'
-    compile 'org.apache.logging.log4j:log4j-core:[2.17.2,3.0)'
+    //compile 'org.apache.logging.log4j:log4j-1.2-api:[2.17.2,3.0)'
+    //compile 'org.apache.logging.log4j:log4j-api:[2.17.2,3.0)'
+    //compile 'org.apache.logging.log4j:log4j-core:[2.17.2,3.0)'
+    //compile 'org.apache.logging.log4j:log4j:[2.17.2,3.0)'
+
+    // log4j 2.18+ suffers from a class loading issue: technical debt
+    compile 'org.apache.logging.log4j:log4j-1.2-api:2.17.2'
+    compile 'org.apache.logging.log4j:log4j-api:2.17.2'
+    compile 'org.apache.logging.log4j:log4j-core:2.17.2'
+    compile 'org.apache.logging.log4j:log4j:2.17.2'
     
     compile 'javax.servlet:javax.servlet-api:3.1.0'
     compile 'org.json:json:20160212'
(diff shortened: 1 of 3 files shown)


cadc-util/build.gradle
```diff
@@ -16,15 +16,22 @@ sourceCompatibility = 1.8
...
+version = '1.7.3'
 
 description = 'OpenCADC core utility library'
 def git_url = 'https://github.com/opencadc/core'
 
 dependencies {
-    compile 'org.apache.logging.log4j:log4j-1.2-api:[2.17.2,3.0)'
-    compile 'org.apache.logging.log4j:log4j-api:[2.17.2,3.0)'
-    compile 'org.apache.logging.log4j:log4j-core:[2.17.2,3.0)'
+    //compile 'org.apache.logging.log4j:log4j-1.2-api:[2.17.2,3.0)'
```
hjeeves, line 25: You've left the original (in the end correct?) compile statements in here - it'll be worth adding a comment about why with the line below. It might be obvious, but might not.

cadc-util/src/main/java/ca/nrc/cadc/net/AuthChallenge.java
```diff
@@ -0,0 +1,145 @@
+/*
+************************************************************************
+*******************  CANADIAN ASTRONOMY DATA CENTRE  *******************
+**************  CENTRE CANADIEN DE DONNÉES ASTRONOMIQUES  **************
+*
+*  (c) 2020.                            (c) 2020.
```
hjeeves, line 6: (c) 2022

cadc-util/src/main/java/ca/nrc/cadc/net/HttpPost.java
```diff
@@ -401,9 +401,12 @@ private void doPost(HttpURLConnection conn, Map<String,List<Object>> params, Map
                 } else {
                     sb.append(pe.getKey());
                     sb.append("=");
+                    if (v instanceof char[]) {
+                        v = String.valueOf((char[]) v);
+                    }
                     sb.append(URLEncoder.encode(v.toString(), "UTF-8"));
-                    sb.append("&");
                 }
+                sb.append("&");
```
hjeeves, line 409: Is this okay here considering the LINE_FEED adds in the 'if' part of this stanza?

cadc-util/src/main/java/ca/nrc/cadc/net/AuthChallenge.java
```diff
@@ -0,0 +1,145 @@
...
+     * Parse a www-authenticate challenge of the form
+     * name [key="value"[, key="value" ...]]
+     * 
+     */
+    public AuthChallenge(String challenge) {
+        String s = challenge.replace(",", " "); // remove comma(s) between k=v pairs
+
+        String[] parts = s.split("[ \t]");
+        this.name = parts[0];
+        if (parts.length == 1) {
```
hjeeves, line 96: can parts.length be < 1?

hjeeves (changes requested):
There's a couple of comments.

cadc-util/build.gradle
pdowler: I put the explanation with the currently active dependencies (hopefully temporary); it's implicit that it apllies to the commented out "normal" case because you can't have duplicates

cadc-util/src/main/java/ca/nrc/cadc/net/AuthChallenge.java
pdowler: doh, templates out of date because of working from home since 2020 :-)

cadc-util/src/main/java/ca/nrc/cadc/net/HttpPost.java
pdowler: good catch... fixed. It used to be inside the else

cadc-util/src/main/java/ca/nrc/cadc/net/AuthChallenge.java
pdowler: if the input string is not null, there is at least one part.

if it is null: NPE from the first line of the ctor, which I'm ok with at this point

PR #201, opened 2022-10-26
tweaks to authorization header handling

pdowler:
so other challenge types can be supported in custom Authenticator

### DIFF
--- a/cadc-util/src/main/java/ca/nrc/cadc/auth/AuthenticationUtil.java
+++ b/cadc-util/src/main/java/ca/nrc/cadc/auth/AuthenticationUtil.java
@@ -350,6 +350,10 @@ public class AuthenticationUtil {
 
         Subject subject = new Subject(false, principals, publicCred, privateCred);
         subject = validateSubject(subject);
+        // remove all un-validated authorization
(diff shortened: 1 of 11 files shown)


cadc-util/src/main/java/ca/nrc/cadc/auth/AuthenticationUtil.java
```diff
@@ -350,6 +350,10 @@ public static Subject getSubject(PrincipalExtractor principalExtractor, boolean
 
         Subject subject = new Subject(false, principals, publicCred, privateCred);
         subject = validateSubject(subject);
+        // remove all un-validated authorization
+        for (AuthorizationTokenPrincipal tp : subject.getPrincipals(AuthorizationTokenPrincipal.class)) {
+            subject.getPrincipals().remove(tp);
+        }
```
andamian, line 356: I don't understand the reason here.

pdowler: if the validate() step above did not process an AuthorizationTokenPrincipal (into an AuthorizationToken) it needs to be dropped because it is a raw input from user (via authorization header) and unknown to the server side code... it may break downstream code that looks at the set of principals because it will be an unrecognized type of principal.

just a missing safety measure needed because of change in ServletPrincipalExtractor.

PR #204, opened 2022-11-22
cadc-util: temporary support for both incorrect and correct digest he…

pdowler:
…ader parsing

### DIFF
--- a/cadc-util/src/main/java/ca/nrc/cadc/net/DigestUtil.java
+++ b/cadc-util/src/main/java/ca/nrc/cadc/net/DigestUtil.java
@@ -69,6 +69,7 @@
 
 package ca.nrc.cadc.net;
 
+import ca.nrc.cadc.util.HexUtil;
 import ca.nrc.cadc.util.StringUtil;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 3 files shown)


cadc-util/src/main/java/ca/nrc/cadc/net/DigestUtil.java
```diff
@@ -200,11 +201,18 @@ public static String base64Encode(String value) {
      * @param value Value to decode.
      * @return Base64 decode of the value.
      */
-    public static String base64Decode(String value) {
+    public static String base64Decode(String scheme, String value) {
```
andamian, line 204: Change in API - hopefully no other packages are using this directly. I did a quick check locally and that seems to be the case.

cadc-util/src/main/java/ca/nrc/cadc/net/DigestUtil.java
```diff
@@ -200,11 +201,18 @@ public static String base64Encode(String value) {
...
         if (value == null) {
             throw new IllegalArgumentException("value can not be null");
         }
         byte[] decoded = Base64.getDecoder().decode(value.getBytes(StandardCharsets.UTF_8));
+
+        // temporary spec-compliant support for md5
```
andamian, line 210: I would have put the incorrect code in the temporary `if` block to be easily removed later

andamian (approved):

cadc-util/src/main/java/ca/nrc/cadc/net/DigestUtil.java
pdowler: yes, in my full change to correctness I actually made these methods private. API methods that take `String` and have vague javadocs are really dangerous and should be kept private where possible.

anyway, private didn't seem to break anything.

PR #212, opened 2023-01-20
Closes #211

### LINKED ISSUE #211: Allow anonymous access to log control if desired
Some systems do not have access control around the log control servlet.  It would be nice to provide a way to allow anonymous access, or at least have some known token that can be used.  The current library requires authentication.

at88mph:


at88mph:
Fixes #211

pdowler (commented):
This is super dangerous as bumping logs to debug can easily be a used as a DoS attack.
Also, bumping them the other way to warn or error (from info) can hide loggable requests.

So I'm not very keen on adding support for things people should never enable. How about we close this PR and discuss options for the use case in the issue??

at88mph:
Sure!

PR #213, opened 2023-01-27
Alt auth access

at88mph:
Allow for BASIC authentication (username/password) with `username` key specified in the log properties file.
Allow for Secret key access with `secret` key specified in the log properties file.  Can be set as a request parameter with `?secret=` or as a request header (`x-cadc-logcontrol`).

### DIFF
--- a/cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
+++ b/cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
@@ -71,6 +71,7 @@ package ca.nrc.cadc.log;
 
 import ca.nrc.cadc.auth.AuthenticationUtil;
 import ca.nrc.cadc.auth.Authorizer;
+import ca.nrc.cadc.auth.HttpPrincipal;
 import ca.nrc.cadc.cred.client.CredUtil;
 import ca.nrc.cadc.net.TransientException;
 import ca.nrc.cadc.util.Log4jInit;
@@ -167,6 +168,9 @@ public class LogControlServlet extends HttpServlet {
     private static final String LOG_CONTROL_PROPERTIES = "logControlProperties";
     static final String USER_DNS_PROPERTY = "user";
     static final String GROUP_URIS_PROPERTY = "group";
+    static final String USERNAME_PROPERTY = "username";
+    static final String SECRET_PROPERTY = "secret";
+    static final String REQUEST_HEADER_SECRET_KEY = "x-cadc-logcontrol";
 
     private Level level = null;
     private List<String> packages;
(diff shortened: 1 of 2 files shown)


cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -167,6 +168,9 @@ public class LogControlServlet extends HttpServlet {
     private static final String LOG_CONTROL_PROPERTIES = "logControlProperties";
     static final String USER_DNS_PROPERTY = "user";
     static final String GROUP_URIS_PROPERTY = "group";
+    static final String USERNAME_PROPERTY = "username";
+    static final String SECRET_PROPERTY = "secret";
+    static final String REQUEST_HEADER_SECRET_KEY = "x-cadc-logcontrol";
```
pdowler, line 173: I was thinking only the secret param and not a custom header (just one way to do it).

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -366,6 +370,24 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+        // We could condense this and just redirect to one of them, but it seems more proper to just carry forward what
+        // the caller submitted.
+        String secretQuery = request.getParameter(SECRET_PROPERTY);
+        String secretHeader = request.getHeader(REQUEST_HEADER_SECRET_KEY);
+
+        if (StringUtil.hasLength(secretQuery)) {
+            url = url + "?" + SECRET_PROPERTY + "=" + secretQuery;
+        }
+
+        if (StringUtil.hasLength(secretHeader)) {
```
pdowler, line 383: I'm OK with the secret param so that `curl --location` works as expected.

The authorization should definitely *not* be set in the response.(it is a request header only).

pdowler (changes requested):

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
at88mph: OK.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
at88mph: That was a dumb thing to do!  Fixed now.

cadc-log/src/main/java/ca/nrc/cadc/log/LogControlServlet.java
```diff
@@ -366,6 +369,17 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
         String url = request.getRequestURI();
+
+        String secretQuery = request.getParameter(SECRET_PROPERTY);
+
+        if (StringUtil.hasLength(secretQuery)) {
+            url = url + "?" + SECRET_PROPERTY + "=" + secretQuery;
+        }
+
+        if (StringUtil.hasLength(request.getHeader("Authorization"))) {
+            response.setHeader("Authorization", request.getHeader("Authorization"));
```
pdowler, line 380: this still needs to be removed

pdowler (changes requested):

pdowler (approved):

PR #214, opened 2023-02-08
Authenticator and IdentityManager API and load changes

pdowler:


cadc-util/src/main/java/ca/nrc/cadc/auth/AuthenticationUtil.java
```diff
@@ -138,76 +134,70 @@ public class AuthenticationUtil {
...
-            c = Class.forName(cname);
-            Object o = c.newInstance();
-            IdentityManager ret = (IdentityManager) o;
-            log.debug("IdentityManager: " + cname);
-            return ret;
-        } catch (Throwable t) {
-            if (!defaultImplClass.equals(cname) || c != null) {
-                log.error("failed to load configured IdentityManager: " + cname, t);
+        // default
+        return new NoOpIdentityManager();
```
brianmajor, line 161: Maybe add a few log.debug() lines here to indicate which identity manager is being used?

cadc-util/src/main/java/ca/nrc/cadc/auth/AuthenticationUtil.java
```diff
@@ -138,76 +134,70 @@ public class AuthenticationUtil {
...
+    }
+
+    // backwards compat
+    private static Authenticator getAuthenticator(IdentityManager im) {
+        String cname = System.getProperty(Authenticator.class.getName());
+        if (cname != null && im instanceof NoOpIdentityManager) {
```
brianmajor, line 167: Why the check for instanceof NoOpIdentityManager?

brianmajor (changes requested):

cadc-util/src/main/java/ca/nrc/cadc/auth/AuthenticationUtil.java
pdowler: this is a fall back in case they have an old-style Authenticator configured via system prop, but I only wanted to do that if they haven't configured an IdentityManager. If for some reason someone has both configured, it should stick with the IdentityManager.... this will get removed once we get passed any bumpy release.

cadc-util/src/main/java/ca/nrc/cadc/auth/AuthenticationUtil.java
pdowler: will do

pdowler: improved

cadc-util/src/main/java/ca/nrc/cadc/auth/AuthenticationUtil.java
pdowler: I only want this backwards compat to come into play if they have not configured an IdentityManager and have not explicitly configured the NoOpIdentityManager.

Rubin configures their old Authenticator using a java system prop, so this allows that to keep working. They'll have other issues with the IdentityManager interface change, but that can't be avoided without a lot of complexity

PR #216, opened 2023-02-15
fix http digest to encode/decode according to spec

pdowler:


cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
```diff
@@ -657,7 +657,7 @@ public void setReadTimeout(int readTimeout) {
      */
     public void setDigest(URI checksumURI) {
         String algorithm = checksumURI.getScheme();
-        String checksum = DigestUtil.base64Encode(checksumURI.getSchemeSpecificPart());
+        String checksum = DigestUtil.base64EncodeHex(checksumURI.getSchemeSpecificPart());
```
andamian, line 660: Wouldn't `String checksum = DigestUtil.toDigest(checksumURI);` work here?

andamian (approved):

cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
pdowler: you mean move this code into DigestUtil so it provides both `getURI(String)` and `toDigest(URI)`? yes, that would be an improvement in clarity. maybe `getDigestHeader(URI)` to be explicit

PR #225, opened 2023-05-23
cadc-util-1.9.7: use bouncycastle to read certificates

pdowler:


pdowler:
introduces pain in cdp.git that I don't have time to fix

PR #234, opened 2023-10-31
fix error handling so HttpUpload.prepare() will actually throw exceptions; resolve #209

### LINKED ISSUE #209: Do not set OpenID www-authenticate header
OpenID has it's own method for obtaining credentials and we should not be setting the OpenID header in www-authenticate.

Remove: `www-authenticate: ivoa_bearer standard_id="ivo://ivoa.net/sso#OpenID", access_url="https://ws-cadc.canfar.net/ac`

ie, this chunk of code: https://github.com/opencadc/core/blob/7481821a9b6e7ccd405fecd5860711f3b6e11c66/cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java#L509

pdowler:


pdowler:
resolves issue #209

PR #236, opened 2023-12-12
Instantiate RSASignatureGenerator and RSASignatureValidator with in m… (CADC-12974)

andamian:
…emory keys (byte arrays)

### DIFF
--- a/cadc-util/src/main/java/ca/nrc/cadc/util/RsaSignatureGenerator.java
+++ b/cadc-util/src/main/java/ca/nrc/cadc/util/RsaSignatureGenerator.java
@@ -96,12 +96,10 @@ import org.apache.log4j.Logger;
  * This class is used to sign and/or verify signed messages. The class requires
  * and RSA private key to sign a message.
  * 
- * <p>The key is passed to the class via the MessageRSA.keys file in the 
- * classpath. This class cannot be instantiated without this file containing
- * a private RSA key.
+ * <p>The key is passed to the class via a file or as a binary array.
  * 
  * <p>Format of the key:
- * The private key in the MessageRSA.keys file must be in PEM TKCS#8 to work 
+ * The private key in the MessageRSA.keys file must be in PEM TKCS#8 to work
  * with basic Java. These keys are in text format delimited by the following 
  * rows: 
  * "-----BEGIN PRIVATE KEY-----" and "-----END PRIVATE KEY-----".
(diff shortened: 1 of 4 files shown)


cadc-util/src/main/java/ca/nrc/cadc/util/RsaSignatureVerifier.java
```diff
@@ -165,6 +163,15 @@ public RsaSignatureVerifier(File keyFile) {
         }
         init(keyFile);
     }
+
+    public RsaSignatureVerifier(byte[] keyContent) {
+        init(keyContent);
+    }
+
+    public RsaSignatureVerifier(byte[] keyContent, boolean isPrivateKey) {
```
pdowler, line 171: I don't really grok what these ctor with `boolean  isPrivateKey` arg are for... if that arg is true, there is no init.

cadc-util/src/main/java/ca/nrc/cadc/util/RsaSignatureGenerator.java
```diff
@@ -150,86 +148,94 @@ public RsaSignatureGenerator(File keyFile) {
         super(keyFile, true);
         initPrivateKey(keyFile);
     }
+
```
pdowler, line 151: I think RsaSignatureGenerator needs a method like
```
public static KeyPair generateKeyPair(int len)
```

It's essentially the first half of the current generateKeyPair(File, File, int) method. The code in raven and vault can generate in-memory keys when necessary; cavern needs the existing File based method. I'm assuming byte correct `byte[]` is available via KeyPair methods (looks like it).

cadc-util/src/test/java/ca/nrc/cadc/util/RSASignatureGeneratorValidatorTest.java
```diff
@@ -77,6 +80,30 @@ public void testSignVerify() throws Exception {
         }
     }
 
+    @Test
+    public void testSignVerifyBytes() throws Exception {
+        final String testString = "cadcauthtest1-" + new Date();
```
pdowler, line 85: That new static method above would be used here (inside for loop) and encapsulate some of this code

pdowler (changes requested):

cadc-util/src/main/java/ca/nrc/cadc/util/RsaSignatureVerifier.java
andamian: This happens because the default deprecated actor requires the key file to be presented. Unfortunately, this ctor is still used in a few places. I could clean it but it requires other package releases. It looks ugly but I couldn't find a better solution. In the future, we need to change checkstyle to fail on deprecated code.

PR #242, opened 2024-02-14
Entity metaChecksum algorithm

pdowler:


cadc-util/src/test/java/org/opencadc/persist/EntityTest.java
```diff
@@ -85,7 +85,7 @@ public class EntityTest {
     private static final Logger log = Logger.getLogger(EntityTest.class);
 
     static {
-        Log4jInit.setLevel("org.opencadc.persist", Level.INFO);
+        Log4jInit.setLevel("org.opencadc.persist", Level.DEBUG);
```
andamian, line 88: Might want to turn it back to `INFO`?

cadc-util/src/test/java/org/opencadc/persist/EntityTest.java
```diff
@@ -105,9 +105,28 @@ public void testTemplate() {
...
+    public void testEntityTruncateDates() {
+        // the caom2-2.4 configuration
+        doEntityTest(true, false);
+        doNewVersionTest(true, false);
+    }
+    
+    @Test
+    public void testEntityDigestFieldNames() {
+        // the cadc-vos-2.x configuration
+        //doEntityTest(false, true);
```
andamian, line 123: This is not working?

cadc-util/src/test/java/org/opencadc/persist/EntityTest.java
```diff
@@ -105,9 +105,28 @@ public void testTemplate() {
...
+        doNewVersionTest(false, true);
+    }
+
```
andamian, line 126: Where's the true, true case?

cadc-util/src/test/java/org/opencadc/persist/EntityTest.java
```diff
@@ -105,9 +105,28 @@ public void testTemplate() {
```
andamian, line 105: Is the above comment a placeholder for something?

andamian (approved):

cadc-util/src/test/java/org/opencadc/persist/EntityTest.java
pdowler: oops, I disabled it while debugging the doNewVersionTest below it

cadc-util/src/main/java/org/opencadc/persist/Entity.java
```diff
@@ -232,35 +275,41 @@ public URI computeMetaChecksum(MessageDigest digest) {
...
         try {
             if (o instanceof Entity) {
                 Entity ce = (Entity) o;
-                digest.update(primitiveValueToBytes(ce.id, "Entity.id", digest.getAlgorithm()));
+                digest.update(primitiveValueToBytes(ce.id, "Entity.id"));
                 if (ce.metaProducer != null) {
-                    digest.update(primitiveValueToBytes(ce.metaProducer, "Entity.metaProducer", digest.getAlgorithm()));
+                    digest.update(primitiveValueToBytes(ce.metaProducer, "Entity.metaProducer"));
+                    if (digestFieldNames) {
+                        digest.update(primitiveValueToBytes("metaProducer", "Entity.metaProducer"));
```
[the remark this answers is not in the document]
pdowler: good catch

cadc-util/src/main/java/org/opencadc/persist/Entity.java
```diff
@@ -91,7 +91,12 @@
 import org.apache.log4j.Logger;
 
 /**
- * Base class for entity persistence.
+ * Base class for entity persistence. The metaChecksum algorithm implemented here has a
```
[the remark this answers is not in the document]
pdowler: The default to false is just in the deprecated ctors for backwards compatibiity - in case we did a rebuild of something from inventory or caom2 and it slipped past tests. 

It's deprecated so new usage is encouraged to use the explicit ctor(s) instead.

cadc-util/src/test/java/org/opencadc/persist/EntityTest.java
pdowler: DEBUG is OK here; Entity has a static boolean that actually controls the copious DEBUG logging. I will move that to the `static` block so devs can easily turn it on or off.

cadc-util/src/test/java/org/opencadc/persist/EntityTest.java
pdowler: hmmm, we don't do that, but I guess it's the safest and most general. will add.

cadc-util/src/test/java/org/opencadc/persist/EntityTest.java
pdowler: yeah, just easy to cut&paste for new tests... it's in my junit template in my IDE

cadc-util/src/main/java/org/opencadc/persist/Entity.java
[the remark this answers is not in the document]
andamian: I understand that, but I think is better to document the bug free functionality of the class  (name digest on) and mention the consequences of the flag being turned off.

PR #244, opened 2024-02-22
update cert reading from file to use current bouncycastle

pdowler:
jumped to 1.11 so that cadc-cdp-server and cred service can stay on 1.10.x for awhile

### DIFF
--- a/cadc-util/src/intTest/java/ca/nrc/cadc/auth/SSLUtilTest.java
+++ b/cadc-util/src/intTest/java/ca/nrc/cadc/auth/SSLUtilTest.java
@@ -107,62 +107,6 @@ public class SSLUtilTest
     private static String TEST_PEM_FN = "proxy.pem";
     private static File SSL_PEM;
 
-    private static final String KEY_512 =
-        "-----BEGIN RSA PRIVATE KEY-----\n" +
-        "MIIBOgIBAAJBAOvm3yk/tr7/8ZaT584T54tOviYIpoWWRfwDgd176c0kTfTj43+C\n" +
-        "BgxFcequf5mY51mgD7v38krRA3+xXi/igfsCAwEAAQJBAMqVrQXGcpDaScVPZV1j\n" +
-        "WJAY4lDVUvQb1iQTev4SwPjqUy8H/f0Zt+Bezwf1LaxcHcCFA6QnDxHw6l99/5zw\n" +
-        "p7kCIQD+4rfjcZyYUKwF0C2deKEgvZUjpiLYVyh/G4qKfT2sPwIhAOzu598CHLLn\n" +
-        "LSZoBRJtjuhAr1zUrfkoBsNHQwTKi6tFAiBOpKtyXPKhOHrrTEFWzgqBLJ2gozkr\n" +
-        "ITFYjqnfcycdRwIgbMW1L31hvYRCBxrEEVS4wclIeJ6vC+6jRC1ICEAQZN0CIFe+\n" +
-        "Az22zN/URBRVBK32tI2axHy/j80Asysh+hxalp1F\n" +
-        "-----END RSA PRIVATE KEY-----\n";
-    private static final String KEY_1024 =
-        "-----BEGIN RSA PRIVATE KEY-----\n" +
-        "MIICXQIBAAKBgQDIcNIXiBAPuOjTNWJUbsI+TR1FSXNGM2bv2naU7pKQ0OWI+DDs\n" +
-        "K1xlctgTi5WrfsMjPKoM+0zpVT5qjUrHyFatbTu9tYjLblPmi/yzOEOIloqZ8lF1\n" +
-        "hrkUG98f8IBbgx4BbkXscVKdP9awngEpIYrZt3QXLUwUP+oF2PCGH5f+nwIDAQAB\n" +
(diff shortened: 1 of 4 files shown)


cadc-util/src/main/java/ca/nrc/cadc/auth/SSLUtil.java
```diff
@@ -618,35 +547,73 @@ public static X509CertificateChain readPemCertificateAndKey(File pemFile)
...
+                    byteSize += po.getContent().length;
+                    log.debug(" certificate: " + po.getContent().length);
+                } else if ("rsa private key".equalsIgnoreCase(po.getType())) {
+                    KeyFactory keyFactory = KeyFactory.getInstance("RSA");
+                    PKCS8EncodedKeySpec keySpec = new PKCS8EncodedKeySpec(po.getContent());
+                    privateKey = keyFactory.generatePrivate(keySpec);
+                    log.debug(" private key: " + po.getContent().length);
+                } else {
+                    log.warn("readPEM: unexpected PemObject type: " + po.getType() + " aka " + obj.getClass().getName());
+                }
```
andamian, line 597: Didn't know of all these types that can be stored in PEM. Do we need them all?

cadc-util/src/main/java/ca/nrc/cadc/auth/SSLUtil.java
```diff
@@ -618,35 +547,73 @@ public static X509CertificateChain readPemCertificateAndKey(File pemFile)
...
+            }
+            
+            obj = parser.readPemObject();
+        }
+
```
andamian, line 602: Should it check here if the private key was found?

cadc-util/src/main/java/ca/nrc/cadc/auth/SSLUtil.java
```diff
@@ -618,35 +547,73 @@ public static X509CertificateChain readPemCertificateAndKey(File pemFile)
...
+            
 
-    /**
-     * Parses a byte array and constructs the corresponding RSAPrivateCrtKeySpec.
-     * 
-     * @param code byte array containing the key
-     * @return RSAPrivateCrtKeySpec
-     * @throws IOException
-     */
+    /*
```
andamian, line 616: Need it anymore?

andamian (approved):

cadc-util/src/main/java/ca/nrc/cadc/auth/CertCmdArgUtil.java
```diff
@@ -132,16 +132,6 @@ private static Subject initSubjectByPem(String fnPem, boolean nullOnNotFound) {
         return SSLUtil.createSubject(certKeyFile);
     }
 
-    private static Subject initSubjectByCertKey(String fnCert, String fnKey, boolean nullOnNotFound) {
```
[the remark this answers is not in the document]
pdowler: yes, I think I made them private so I could see where they were used by breaking things; I'll try to do more cleanup

cadc-util/src/intTest/java/ca/nrc/cadc/auth/SSLUtilTest.java
```diff
@@ -225,6 +226,72 @@ public void testGetSocketFactoryFromFile() throws Exception
             Assert.fail("unexpected exception: " + t);
         }
     }
+    
+    @Test
+    public void testReadCert() {
+        try {
+            File f = new File(System.getProperty("user.home") + "/.ssl/" + System.getProperty("user.name") + ".pem");
```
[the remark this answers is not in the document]
pdowler: this is reading a normal cert rather than a proxy cert; part of what I ran into was they had different structures sometimes so I wanted to test both

cadc-util/src/main/java/ca/nrc/cadc/auth/SSLUtil.java
```diff
@@ -290,15 +239,17 @@ public static Subject createSubject(File certKeyFile) {
...
 
-    static byte[] getPrivateKey(byte[] certBuf) throws IOException {
+    @Deprecated
+    private static byte[] getPrivateKey(byte[] certBuf) throws IOException {
         BufferedReader rdr = new BufferedReader(new InputStreamReader(new ByteArrayInputStream(certBuf)));
         String line = rdr.readLine();
         StringBuilder base64 = new StringBuilder();
         while (line != null) {
-            if (line.startsWith("-----BEGIN RSA PRIVATE KEY-")) {
+            if (line.matches("-----BEGIN.*PRIVATE KEY-----")) {
```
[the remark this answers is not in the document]
pdowler: no, that was something I tried but it didn't help. but it's part of the cleanup.

cadc-util/src/main/java/ca/nrc/cadc/auth/SSLUtil.java
pdowler: yes, the BC lib creates different things sometimes just depending on the order, but I had examples that had all of these variants and kept adding things until I stopped getting the warning :-) I don't recall how one runs into a PrivateKeyInfo on it's own... but this is the main reading code now

cadc-util/src/main/java/ca/nrc/cadc/auth/SSLUtil.java
pdowler: no, it should be able to read a file with only certificates - we don't do it but it is a thing in some 
contexts

PR #249, opened 2024-07-10
Modified OpenIdPrincipal to work with different issuers

andamian:


cadc-util/src/main/java/ca/nrc/cadc/auth/OpenIdPrincipal.java
```diff
@@ -35,33 +35,44 @@
...
     private String openID;
+    private URL issuer;
 
     /**
      * Ctor
      * 
      * @param openID
      */
-    public OpenIdPrincipal(final String openID) {
+    public OpenIdPrincipal(final URL issuer, final String openID) {
```
pdowler, line 56: javadoc: should have the issuer param

Should we make it more clear in the field name `openID` that this is the `sub` (subject) field?

pdowler (changes requested):

pdowler (approved):

PR #256, opened 2025-01-20
CADC-13577 add list() method to KeyValueDAO

jburke-cadc:


pdowler (approved):

cadc-util/src/main/java/ca/nrc/cadc/db/version/KeyValueDAO.java
```diff
@@ -197,6 +198,41 @@ public void delete(String name) {
         jdbc.update(sql, arg);
     }
 
+    public List<KeyValue> list() {
+        String select = String.format("SELECT %s, %s, %s FROM %s", columnNames[0], columnNames[1], columnNames[2], tableName);
```
pdowler, line 202: The `select` SQL is almost a duplicate of the sql generated inside the SelectStatementCreator; if you made the WHERE (name) optional, the SQL generation and probably the whole SelectStatementCreator could be re-used.

cadc-util/src/main/java/ca/nrc/cadc/db/version/KeyValueDAO.java
```diff
@@ -197,6 +198,41 @@ public void delete(String name) {
...
+        ResultSetExtractor<List<KeyValue>> keyValueListExtractor = new KeyValueListExtractor();
+        List<KeyValue> keyValues;
+        try {
+            keyValues = jdbc.query(select, keyValueListExtractor);
+        } catch (Exception ex) {
+            Connection con = null;
+            try {
+                log.debug("query fail - check if table exists: " + tableName);
```
pdowler, line 210: Now that we are intending to use this for more than just the InitDatabase + ModelVersionDAO use case, this code to check for table existence and return null for various reasons) isn't that great anymore.

Better plan:
1. write a single method that checks for table existence and use that explicitly in InitDatabase; in all other cases table not existing is an error
2. allow other methods to throw spring-jdbc runtime exception but document that in the javadocs

cadc-util/src/main/java/ca/nrc/cadc/db/version/KeyValueDAO.java
```diff
@@ -311,6 +347,22 @@ public Object extractData(ResultSet rs) throws SQLException {
         }
     }
 
+    private class KeyValueListExtractor implements ResultSetExtractor<List<KeyValue>> {
+
```
pdowler, line 351: This could/should be a RowMapper instead of a ResultSetExtractor. Then it could be used in the list() method and in the (poorly named) ModelVersionExtractor, which should be renamed to KeyValueExtractor. The reason that one is a ResultSetExtractor is that it simplifies the get() and lock() methods by returning a single object. RowMapper is generally quite re-usable.

pdowler (changes requested):
The existing intTest (InitDatabaseTest) tested ModelVersion and thus KeyValue as well, but  it does not use the new `list()` method.

I'd probably suggest a separate KeyValueDAOTest that uses all the methods (put, get, delete, list).

cadc-util/src/main/java/ca/nrc/cadc/db/version/KeyValueDAO.java
pdowler: also, RowMapper can be used with `queryForObject` for a type-safe single-row return value; ResultSetExtractor is only needed if the returned object spans multiple rows in the result set.

cadc-util/src/main/java/ca/nrc/cadc/db/version/KeyValueDAO.java
jburke-cadc: Used the SelectStatementCreator for both get() and list() methods.

cadc-util/src/main/java/ca/nrc/cadc/db/version/KeyValueDAO.java
jburke-cadc: Updated the KeyValueDAO to allow the methods to throw the Spring exception on error, and check if the table exists in InitDatabase.

cadc-util/src/main/java/ca/nrc/cadc/db/version/KeyValueDAO.java
jburke-cadc: Updated to use a RowMapper.

cadc-util/src/main/java/ca/nrc/cadc/db/version/KeyValueDAO.java
```diff
@@ -197,10 +193,24 @@ public void delete(String name) {
...
+        SelectStatementCreator sel = new SelectStatementCreator();
+        return jdbc.query(sel, new KeyValueRowMapper());
+    }
+
     private class SelectStatementCreator implements PreparedStatementCreator {
 
         private final boolean forUpdate;
         private String model;
+        private String name;
```
pdowler, line 212: This looks wrong in several ways.

1. setWhere() is never called and should be removed;
2. the `name` column in the db is the only query constraint (when not listing)
3. listing (model == null) and locking (`forUpdate`) should never be used together

There is no need for both `model` and `name` fields... `model` makes less sense now so should be dropped in favour of `name` (eg in `setValue(String name)` and in `loadValues()`). 

The conditional in creating SQL would then be correct, but the SQL itself would be wrong: the column name is literally "name", not the value of name.

for 3, we need a single method that takes the `name` and `forUpdate` value... so that implies something like `setValue(String, boolean)`

cadc-util/src/intTest/java/ca/nrc/cadc/db/version/KeyValueDAOTest.java
```diff
@@ -0,0 +1,210 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+public class KeyValueDAOTest {
+    private static final Logger log = Logger.getLogger(KeyValueDAOTest.class);
+
+    private static final String SCHEMA = "dbversion";
+    private static final String DATABASE = "cadctest";
+    private static final String MODELNAME = "KeyValueDAOTest";
```
pdowler, line 93: these are just key name and key value? "MODEL" is just misleading here... and not actually used.

In CREATE constant below, should use TABLE_NAME

cadc-util/src/intTest/java/ca/nrc/cadc/db/version/KeyValueDAOTest.java
```diff
@@ -0,0 +1,210 @@
...
+            "(name varchar(32) not null primary key,value varchar(32) not null,lastModified timestamp not null)";
+
+    public KeyValueDAOTest() {
+    }
+
+    @BeforeClass
+    public static void setup() {
+        Log4jInit.setLevel("ca.nrc.cadc.db.version", Level.DEBUG);
+
+        try {
```
pdowler, line 108: setup( ) methods don't need to `assert` anything, just throw exceptions if they fail

cadc-util/src/intTest/java/ca/nrc/cadc/db/version/KeyValueDAOTest.java
```diff
@@ -0,0 +1,210 @@
...
+            expected.value = "test-value";
+            dao.put(expected);
+
+            KeyValue actual = dao.get(expected.getName());
+            Assert.assertNotNull(actual);
+            Assert.assertEquals(expected.getName(), actual.getName());
+            Assert.assertEquals(expected.value, actual.value);
+            Assert.assertEquals(expected.lastModified, actual.lastModified);
+
+            try {
```
pdowler, line 161: Why this extra try/catch?

cadc-util/src/intTest/java/ca/nrc/cadc/db/version/KeyValueDAOTest.java
```diff
@@ -0,0 +1,210 @@
...
+            KeyValue kv2 = new KeyValue("test-key-2");
+            kv2.value = "test-value-2";
+            dao.put(kv2);
+
+            KeyValue kv3 = new KeyValue("test-key-3");
+            kv3.value = "test-value-3";
+            dao.put(kv3);
+
+            List<KeyValue> actual = null;
+            try {
```
pdowler, line 190: Why this extra try/catch?

pdowler (changes requested):

cadc-util/src/main/java/ca/nrc/cadc/db/version/KeyValueDAO.java
jburke-cadc: Updated for the above comments.

cadc-util/src/intTest/java/ca/nrc/cadc/db/version/KeyValueDAOTest.java
jburke-cadc: Yes poorly named and not used either, removed.

cadc-util/src/intTest/java/ca/nrc/cadc/db/version/KeyValueDAOTest.java
jburke-cadc: Updated to throw an exceptions encountered during setup.

cadc-util/src/intTest/java/ca/nrc/cadc/db/version/KeyValueDAOTest.java
jburke-cadc: I guess to give a better error message about where the failure occurred, but it could've failed multiple times before then, so likely not terribly helpful. Removed.

cadc-util/src/intTest/java/ca/nrc/cadc/db/version/KeyValueDAOTest.java
jburke-cadc: Same as above.

pdowler (approved):

PR #267, opened 2025-04-24
add code to augment ivoa challenge headers

pdowler:


cadc-rest/src/main/java/ca/nrc/cadc/rest/RestServlet.java
```diff
@@ -484,25 +484,62 @@ static void setAuthenticateHeaders(Subject subject, String realm, SyncOutput out
...
+                            || challenge.equals(AuthenticationUtil.CHALLENGE_TYPE_IVOA_COOKIE)) {
+                        final URI standardID;
+                        if (challenge.equals(AuthenticationUtil.CHALLENGE_TYPE_IVOA_X509)) {
+                            standardID = Standards.CRED_PROXY_10;
+                        } else {
+                            // make something up for now
+                            standardID = URI.create("ivo://ivoa.net/std/Cred#cookie-issuer");
+                        }
+                        try {
+                            URI certGenService = localAuthority.getServiceURI(standardID.toASCIIString());
```
brianmajor, line 511: `credGenService` instead of `certGenService` in anticipation of the cookie-issuer service coming online?
Relatedly, I don't think lines in the block below (512 down) will be correct if it does come online.  So there's only a partial step towards a working cookie issuing service.  Just need to remember this code will need updating.

brianmajor (approved):

PR #268, opened 2025-06-12
cadc-rest: capture multiple inline content objects with same name

pdowler:
content type automatically upgraded from Object to List

bug: previously only retained the last object processed and dropped the others

### DIFF
--- a/cadc-rest/src/main/java/ca/nrc/cadc/rest/SyncInput.java
+++ b/cadc-rest/src/main/java/ca/nrc/cadc/rest/SyncInput.java
@@ -120,6 +120,12 @@ public class SyncInput {
             }
         }
     }
+    
+    // for subclass stubs used in unit tests
+    protected SyncInput() {
+        this.request = null;
+        this.inlineContentHandler = null;
+    }
 
     /**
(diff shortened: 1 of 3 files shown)


cadc-rest/src/main/java/ca/nrc/cadc/rest/SyncInput.java
```diff
@@ -359,7 +367,19 @@ private void processStream(String name, String contentType, InputStream inputStr
         }
 
         InlineContentHandler.Content c = inlineContentHandler.accept(name, contentType, new CloseWrapper(inputStream));
-        content.put(c.name, c.value);
+        Object o = content.get(c.name);
+        if (o == null) {
+            content.put(c.name, c.value);
+        } else if (o instanceof List) {
+            List cur = (List) o;
+            cur.add(c);
```
andamian, line 375: Is `content.put(c.name, cur);` missing here?

pdowler: no, the list is already a value in the map, so just add another value to list is sufficient.

andamian (approved):

PR #269, opened 2025-06-12
Changes required to support JWT tokens (CADC-14107)

andamian:


cadc-util/src/main/java/ca/nrc/cadc/auth/SignedToken.java
```diff
@@ -248,6 +248,25 @@ private static StringBuilder getContent(SignedToken token) {
         return sb;
     }
 
+    /**
+     * Checks whether a token is a CADC access token or not
```
pdowler, line 252: "a CADC access token" -> SignedToken

also, `Base64.decodeString(String)` exists and could be used instead of `new String`... it's the same code but intended for this use case

should be `catch(IllegalArgumentException ignore)`


Q. Why does this handle plain text and base64-encoded test? And why is the condition different in those two cases? Aren't all SignedToken values in headers base64 encoded?

pdowler (changes requested):

cadc-util/src/main/java/ca/nrc/cadc/auth/SignedToken.java
andamian: Fixed. Can't remember why plain text. I've removed it.

cadc-util/src/main/java/ca/nrc/cadc/auth/TokenValidator.java
```diff
@@ -149,11 +149,14 @@ public static Subject validateTokens(Subject subject) throws NotAuthenticatedExc
                 log.debug("credentials: " + credentials);
 
                 try {
+                    if (!SignedToken.isSignedToken(credentials)) {
```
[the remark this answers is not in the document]
andamian: Indeed - once deprecated code is removed, the logic becomes simpler.

pdowler (changes requested):
It still looks like `isSignedToken` could be easily fooled since it only looks for the delim (=). In `parse(String text)`  it looks for the one required field (EXPIRY_LABEL) so maybe do that to more likely avoid a false positive?

pdowler (approved):

PR #284, opened 2026-04-05
Checkstyle version upgrade 

aratikakadiya:


opencadc.gradle
```diff
@@ -7,14 +7,14 @@ configurations {
 }
 
 dependencies {
-    testImplementation 'com.puppycrawl.tools:checkstyle:8.2'
-    checkstyleDep 'org.opencadc:cadc-quality:[1.0,1.1)'
+    testImplementation 'com.puppycrawl.tools:checkstyle:10.12.5'
+    checkstyleDep 'org.opencadc:cadc-quality:[1.1,)'
```
pdowler: should use an upper bound eg  `[1.1,1.2)` here so we can gradually release an update next time we do an incompat change

pdowler (changes requested):
Can you provide a brief (point form) summary of the kinds of changes that this update requires in code? (eg things that are more strict). It might be worth adding that list to the cadc-quality/README.md (eg have a *Changes in 1.1* section).

For example, in cadc-rest SyncOutput.java I see that some blank lines in the import section were removed... are blank lines not allowed or was that your IDE reformatting? To be clear, I am in favour of not allowing blank lines.

My main question: code reformatting is a pain to review and I don't really want to look at the code itself... I will assume there are no functional changes, but I will probably build and install the two libs locally and publish new versions in a few weeks just in case.

cadc-quality/build.gradle
```diff
@@ -11,19 +11,19 @@ repositories {
 
 apply from: '../opencadc.gradle'
 
-sourceCompatibility = 1.8
+sourceCompatibility = 11
 
 group = 'org.opencadc'
 
-version = '1.0.1'
+version = '1.1'
```
pdowler: need to include path in version, so `1.1.0`

pdowler (changes requested):

pdowler (approved):

PR #285, opened 2026-04-15
update metaChecksum algorithm

pdowler:
digest a single 0 byte after each list item
PrimitiveWrapper by itself and inside collections
handle arrays of primitives, usually from PrimtiveWrapper unwrap use cases: CAOM-2.5 algorithm updates


cadc-util/src/main/java/org/opencadc/persist/Entity.java
```diff
@@ -454,14 +420,32 @@ protected final void calcMetaChecksum(Class c, Object o, MessageDigestWrapper di
                         if (digestFieldNames && num < digest.getNumBytes()) {
                             digest.update(fieldNameToBytes(cf)); // field name
                         }
+                    } else if (ac.isArray()) {
+                        Iterator iter = new ArrayIterator(fo);
+                        if (iter.hasNext()) {
+                            // arrays of primitive only
```
andamian: How is this enforced in general? If not, should a check exist here?

cadc-util/src/main/java/org/opencadc/persist/Entity.java
```diff
@@ -454,14 +420,32 @@ protected final void calcMetaChecksum(Class c, Object o, MessageDigestWrapper di
...
+                            while (iter.hasNext()) {
+                                Object co = iter.next();
+                                Class cc = co.getClass();
```
andamian: No use of `cc`

cadc-util/src/main/java/org/opencadc/persist/Entity.java
```diff
@@ -472,9 +456,23 @@ protected final void calcMetaChecksum(Class c, Object o, MessageDigestWrapper di
...
                                     // depth-first recursion
                                     calcMetaChecksum(cc, co, digest);
+                                } else if (cc.isArray()) {
+                                    // use case: Interval -> Object[]
+                                    Iterator ai = new ArrayIterator(co);
+                                    if (ai.hasNext()) {
+                                        // arrays of primitive only
+                                        while (ai.hasNext()) {
+                                            Object ico = ai.next();
+                                            Class icc = co.getClass();
```
andamian: No use.

cadc-util/src/main/java/org/opencadc/persist/Entity.java
```diff
@@ -677,7 +697,7 @@ protected byte[] primitiveValueToBytes(Object o, String name) {
             return ret;
         }
 
-        throw new UnsupportedOperationException("unexpected primitive/value type: " + o.getClass().getName());
+        throw new UnsupportedOperationException("unexpected primitive/value type: " + o.getClass().getName() + " field: " + name);
```
andamian, line 714: Or maybe this is where arrays of objects are caught.

andamian (approved):

cadc-util/src/main/java/org/opencadc/persist/Entity.java
pdowler: I changed PrimitiveWrapper slightly to document what kinds of things can be returned and I limited arrays to primtives; in the primitiveValueToBytes I explicitly support only double[] and long[] and fail otherwise.

Then this reflective array handling was completely removed. It was only there because Interval in cadc-dali returned Object[] but I fixed that to return one of the supported types.

cadc-util/src/main/java/org/opencadc/persist/Entity.java
pdowler: I think this is also gone now

cadc-util/src/main/java/org/opencadc/persist/Entity.java
pdowler: also gone

cadc-util/src/main/java/org/opencadc/persist/Entity.java
pdowler: yes, this is where unsupported types land

cadc-util/src/main/java/org/opencadc/persist/Entity.java
```diff
@@ -496,6 +482,28 @@ protected final void calcMetaChecksum(Class c, Object o, MessageDigestWrapper di
         }
     }
     
+    private static class ArrayIterator implements Iterator {
```
andamian: Do you still need this?

PR #288, opened 2026-06-16
cadc-util: tweak error handling

pdowler:
add PermissionDeniedException to replace deprecated AccessControlException

### DIFF
--- a/cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
+++ b/cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
@@ -3,7 +3,7 @@
 *******************  CANADIAN ASTRONOMY DATA CENTRE  *******************
 **************  CENTRE CANADIEN DE DONNÉES ASTRONOMIQUES  **************
 *
-*  (c) 2020.                            (c) 2020.
+*  (c) 2026.                            (c) 2026.
 *  Government of Canada                 Gouvernement du Canada
 *  National Research Council            Conseil national de recherches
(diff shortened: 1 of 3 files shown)


cadc-util/src/main/java/ca/nrc/cadc/net/HttpTransfer.java
```diff
@@ -839,26 +835,20 @@ protected void checkTransient(int code, String msg, HttpURLConnection conn)
...
             captureResponseHeaders(conn);
         } catch (SocketTimeoutException ex) {
-            int timeoutRetryDelay = DEFAULT_RETRY_DELAY;
-            if (connectionTimeout > 0 || readTimeout > 0) {
-                // user specified timeouts are indicative of how responsive the client expects the
-                // server to be...
-                timeoutRetryDelay = Math.max(connectionTimeout, readTimeout) / 1000; // ms to sec
-            }
-            throw new TransientException("connection/read timeout: " + url, ex, timeoutRetryDelay);
+            throw new TransientException("url=" + toLoggableString(url) + " connection/read timeout");
```
andamian, line 847: This is not subject to the same retry rules like the other Transient exceptions?

pdowler: it turned out that computing the retry-after from the timeout values made for very long retry loops (because the timeouts can't be that short)... and this is client-side error handling so shouldn't guess at retry-after times... just use the default 1, 2, 4, ... seconds

andamian (approved):

